PR #2, opened 2025-06-03
First PR for compiler

gael-bigot:
- lexing and partial parsing of cairo0 syntax
- compiles expressions, locals and functions

### DIFF
--- a/crates/compiler/src/ast.rs
+++ b/crates/compiler/src/ast.rs
@@ -1,0 +1,581 @@
+use crate::lexer::Token;
+use std::fmt::{self, Debug};
+
+// Types
+
+#[derive(Clone)]
+pub enum Type {
+    Felt,
+    CodeOffset,
+    Pointer(Box<Type>),
+    Pointer2(Box<Type>),
+    Tuple(Vec<Type>),
+    Struct(Identifier),
+    Named(Identifier, Box<Type>),
+    Error,
+}
+
+// Expressions
+
+#[derive(Debug, Clone)]
+pub enum ExprType {
+    IntegerLiteral,
+    Hint,
+    Identifier,
+    Register,
+    Add,
+    Sub,
+    Mul,
+    Div,
+    Pow,
+    Neg,
+    Deref,
+    AddressOf,
+    Cast,
+    New,
+    Eq,
+    Neq,
+    And,
+    FunctionCall,
+    Subscript,
+    TupleOrParen,
+    ErrorExpr,
+}
+
+#[derive(Clone)]
+pub struct Expr {
+    pub token: Option<Token>,
+    pub ident: Option<Identifier>,
+    pub expr_type: ExprType,
+    pub left: Option<Box<Expr>>,
+    pub right: Option<Box<Expr>>,
+    pub type_arg: Option<Type>,
+    pub paren_args: Vec<ExprAssignment>,
+    pub brace_args: Vec<ExprAssignment>,
+}
+
+#[derive(Debug, Clone)]
+pub struct Identifier {
+    pub token: Token,
+}
+
+#[derive(Clone)]
+pub enum ExprAssignment {
+    Expr(Expr),
+    Assign(Identifier, Expr),
+}
+
+// Instructions
+
+#[derive(Clone)]
+pub struct Instruction {
+    pub instruction_type: InstructionType,
+    pub ident: Option<Identifier>,
+    pub args: Vec<Expr>,
+    pub increment_ap: bool,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum InstructionType {
+    CallRel,
+    CallAbs,
+    Call,
+    AssertEq,
+    JmpRel,
+    JmpAbs,
+    Jmp,
+    Jnz,
+    JnzLabel,
+    Ret,
+    AddAp,
+    DataWord,
+}
+
+#[derive(Clone)]
+pub enum CodeElement {
+    Instruction(Instruction),
+    Const,
+    Reference(Identifier, Expr),
+    LocalVar(Identifier, Option<Expr>),
+    TempVar,
+    CompoundAssertEqual(Expr, Expr),
+    StaticAssert,
+    Return(Expr),
+    If(Expr, Vec<CodeElement>, Vec<CodeElement>),
+    FuncCall,
+    Label,
+    Function(Identifier, Vec<Identifier>, Vec<CodeElement>),
+    Struct,
+    NameSpace,
+    TypeDef,
+    WithAttr,
+    With,
+    Hint,
+    Directive,
+    Import,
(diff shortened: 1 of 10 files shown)


crates/compiler/fib.cairo
```diff
@@ -0,0 +1,13 @@
...
+
+func fib(n) {
+    if (n == 0) {
+        return 0;
+    }
+    if (n == 1) {
+        return 1;
+    }
+    return fib(n - 1) + fib(n - 2);
+}
```
ClementWalter, line 13: missing empy line at the end of file (config VS code to add it automatically)

crates/compiler/src/lexer.rs
```diff
@@ -0,0 +1,240 @@
...
+/// let (tokens, error_count) = lex(source, "main.cairo");
+/// assert_eq!(error_count, 0);
+/// ```
+pub fn lex(input: &str, file_name: &str) -> (Vec<Token>, u32) {
+    let mut error_counter = 0;
+    let mut lex = TokenType::lexer(input);
+    let mut tokens = Vec::new();
+    while let Some(token) = lex.next() {
+        let lexeme = lex.slice().to_string();
+        if let Ok(token) = token {
```
zmalatrax, line 122: Consider using `match` to handle the `Result` enum, rather than if/else

crates/compiler/src/ast.rs
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            // Print right child if present
+            if let Some(right) = &self.right {
+                right.fmt_with_indent(f, indent + 1)?;
+                writeln!(f)?;
+            }
+
+            // Print args if present
+            for (_i, arg) in self.paren_args.iter().enumerate() {
```
zmalatrax, line 163: If the iteration count is not used, no need to use `enumerate()`

```suggestion
            for arg in self.paren_args.iter() {
```

crates/compiler/src/lower_to_casm.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub fn compile_expr(&mut self, expr: Expr) -> u64 {
+        match expr.expr_type {
+            ExprType::IntegerLiteral => self.compile_int_literal(expr),
+            ExprType::Add => self.compile_add(expr),
+            ExprType::Sub => self.compile_sub(expr),
+            ExprType::Mul => self.compile_mul(expr),
+            ExprType::FunctionCall => self.compile_function_call(expr),
+            ExprType::Identifier => self.compile_identifier(expr),
+
+            _ => todo!(),
```
zmalatrax, line 213: I'd rather remove the unsupported variant of the `ExprType` enum, and re-introduce them when implementing them

crates/compiler/src/lower_to_casm.rs
```diff
@@ -0,0 +1,462 @@
...
+        self.fp_offset = 0;
+        // counting number of local declarations
+        for code_element in body.clone() {
+            match code_element {
+                CodeElement::LocalVar(_ident, _expr) => {
+                    self.fp_offset += 1;
+                }
+                _ => {}
+            }
+        }
```
zmalatrax, line 241:
```suggestion
        self.fp_offset += body
            .iter()
            .filter(|code_element| matches!(code_element, CodeElement::LocalVar(_, _)))
            .count() as u64;
```

crates/compiler/src/lower_to_casm.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// The list of top-level code elements (functions, statements, etc.) to compile
+    code_elements: Vec<CodeElement>,
+    /// The list of generated CASM instructions
+    casm_instructions: Vec<CasmInstruction>,
+    /// Mapping from local variable names to their frame pointer offsets
+    local_variables: HashMap<String, i32>,
+    /// The number of local variables currently allocated
+    local_variables_count: u64,
+    /// The current frame pointer offset (top of stack)
+    fp_offset: u64,
```
zmalatrax, line 36: why `u64` ?

crates/compiler/src/lower_to_casm.rs
```diff
@@ -0,0 +1,462 @@
...
+            fp_offset: 0,
+            label_counter: 0,
+        }
+    }
+
+    /// Compile all code elements into a vector of CASM instructions.
+    ///
+    /// This is the main entry point for lowering a program.
+    pub fn compile(&mut self) -> Vec<CasmInstruction> {
+        for code_element in self.code_elements.clone() {
```
zmalatrax, line 58: Avoid unnecessary `clone()`

crates/compiler/src/lower_to_casm.rs
```diff
@@ -0,0 +1,462 @@
...
+        let instr = CasmInstruction {
+            instruction_type: CasmInstructionType::MovFpImm,
+            label: None,
+            arg0: self.fp_offset as i32,
+            arg1: expr.token.unwrap().lexeme.parse::<i32>().unwrap(),
+            arg2: 0,
+        };
+        self.casm_instructions.push(instr);
+        self.fp_offset += 1;
+        return self.fp_offset - 1;
```
zmalatrax, line 78: Usually, rust return values are done without the `return` keyword by omitting the semi-colon

```suggestion
        self.fp_offset - 1
```

crates/compiler/src/lower_to_casm.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    /// Compile an addition expression.
+    ///
+    /// Evaluates both operands, emits an add instruction, and returns the result offset.
+    fn compile_add(&mut self, expr: Expr) -> u64 {
+        assert!(matches!(expr.expr_type, ExprType::Add));
+        let left_offset = self.compile_expr(*expr.left.unwrap());
+        let right_offset = self.compile_expr(*expr.right.unwrap());
```
zmalatrax, line 87: Do we want to panic in this case (and other related cases) ?
We should consider adding proper error-handling (with `thiserror`) ;
Must see how it would integrate/conflict with `ariadne` used to display compiler diagnostic errors

crates/compiler/src/lower_to_casm.rs
```diff
@@ -0,0 +1,462 @@
...
+                // jmp <end_label>
+                // <if_label>
+                // <body>
+                // <end_label>
+                let offset = self.compile_expr(Expr::new_binary(
+                    ExprType::Sub,
+                    *expr.left.unwrap(),
+                    *expr.right.unwrap(),
+                ));
+                self.casm_instructions.push(CasmInstruction {
```
zmalatrax, line 340: Consider creating a constructor for `CasmInstruction`, or even helper function to instantiate `CasmInstruction` for each instruction_type like `new_jmp_label_if_neq()`, `new_jmp_label_if_eq`, `new_jmp_label`...

PR #7, opened 2025-06-09
feat(vm): Memory

obatirou:
Closes CORE-740

### DIFF
--- a/crates/runner/src/memory/mod.rs
+++ b/crates/runner/src/memory/mod.rs
@@ -1,0 +1,314 @@
+use num_traits::identities::Zero;
+use stwo_prover::core::fields::m31::M31;
+use stwo_prover::core::fields::qm31::QM31;
+
+/// The maximum number of bits for a memory address, set to 30.
+/// This limits the memory size to 2^30 elements.
+const MAX_MEMORY_SIZE_BITS: u8 = 30;
+
+/// Represents the Cairo VM's memory, a flat, read-write address space.
+///
+/// Memory is addressable by `M31` field elements and stores `QM31` values.
+#[derive(Debug, Clone, Default)]
+pub struct Memory {
+    /// The index of the vector corresponds to the memory address.
+    /// Instructions and data are stored as `QM31` values.
+    pub data: Vec<QM31>,
+}
+
+impl Memory {
+    /// Retrieves a `QM31` value from the specified memory address.
+    ///
+    /// This is used to fetch instructions of the program, which are represented as
+    /// `QM31` values. If the address is out of bounds, it returns
+    /// `QM31::zero()`, simulating that uninitialized memory reads as zero.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `M31` memory address to read from.
+    pub fn get_instruction(&self, addr: M31) -> QM31 {
+        let address = addr.0 as usize;
+        self.data.get(address).copied().unwrap_or_else(QM31::zero)
+    }
+
+    /// Retrieves a value from memory and projects it to a base field element `M31`.
+    ///
+    /// This is used for instruction arguments or other data that are expected to
+    /// be simple field elements. It asserts that the retrieved `QM31` value is
+    /// in the base field (i.e., its extension components are zero).
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `M31` memory address to read from.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value at the address is not a base field element.
+    pub fn get_data(&self, addr: M31) -> M31 {
+        let qm31_value = self.get_instruction(addr);
+        assert!(qm31_value.1.is_zero());
+        assert!(qm31_value.0 .1.is_zero());
+        qm31_value.0 .0
+    }
+
+    /// Checks if a given memory address is within the allowed range (`0` to `2^MAX_MEMORY_SIZE_BITS`).
+    ///
+    /// # Arguments
+    ///
+    /// * `address` - The `M31` address to validate.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the address is out of bounds. This helps prevent memory
+    /// access violations.
+    fn validate_address(address: M31) {
+        if address.0 > (1 << MAX_MEMORY_SIZE_BITS) {
+            panic!("Max memory size is 2 ** {MAX_MEMORY_SIZE_BITS}; got address: {address}.")
+        }
+    }
+
+    /// Inserts a `QM31` value at a specified validated memory address.
+    ///
+    /// If the address is beyond the current memory size, the memory is
+    /// automatically extended and padded with zeros up to the new address.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `M31` memory address to write to.
+    /// * `value` - The `QM31` value to insert.
+    pub fn insert(&mut self, addr: M31, value: QM31) {
+        Self::validate_address(addr);
+        let address = addr.0 as usize;
+
+        // Resize vector if necessary
+        if address >= self.data.len() {
+            self.data.resize(address + 1, QM31::zero());
+        }
+
+        self.data[address] = value;
+    }
+
+    /// Inserts a slice of `QM31` values starting from a given address.
+    ///
+    /// It validates that both the start and end addresses of the slice are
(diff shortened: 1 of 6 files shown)


crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// This is used to fetch instructions of the program, which are represented as
+    /// `QM31` values. If the address is out of bounds, it returns
+    /// `QM31::zero()`, simulating that uninitialized memory reads as zero.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `M31` memory address to read from.
+    pub fn get_instruction(&self, addr: M31) -> QM31 {
+        let address = addr.0 as usize;
+        self.data.get(address).copied().unwrap_or_else(QM31::zero)
```
zmalatrax, line 31: Considering that it is only used to fetch instructions, there shouldn't be a case where the interpreter tries to fetch an instruction to an uninitialized memory cell, no?

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    ///
+    /// * `addr` - The `M31` memory address to read from.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value at the address is not a base field element.
+    pub fn get_data(&self, addr: M31) -> M31 {
+        let qm31_value = self.get_instruction(addr);
+        assert!(qm31_value.1.is_zero());
+        assert!(qm31_value.0 .1.is_zero());
```
zmalatrax, line 50: Make custom errors to handle this rather than panic

crates/runner/src/memory/mod.rs
obatirou: Yes but the rationale was to replicate cairo memory that any read from an uninitialized memory address returns 0 as for now instructions and data share the same memory space

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `address` - The `M31` address to validate.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the address is out of bounds. This helps prevent memory
+    /// access violations.
+    fn validate_address(address: M31) {
```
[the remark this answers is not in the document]
zmalatrax: And it can be marked as const

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        // Check that the entire slice fits within memory limits
+        let start_address = start_addr.0 as usize;
+        let slice_len = values.len();
+        // Since we already checked for empty slice, slice_len >= 1
+        let last_addr = start_addr.0.saturating_add((slice_len - 1) as u32);
+        Self::validate_address(M31::from(last_addr));
+
+        let end_address = start_address + slice_len;
```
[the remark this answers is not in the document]
obatirou: Yes exactly
last_addr is used for validation that the last address written is valid but end_address is for slicing (exclusive)
Will refactor for better readibility

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+        let result = memory.get_instruction(addr);
+        assert_eq!(result, QM31::zero());
+    }
+
+    #[test]
+    fn test_get_data() {
+        let mut memory = Memory::default();
+        let addr = M31::from(42);
+        // Create a QM31 value that represents a base field element (only first component non-zero)
+        let value = QM31::from_m31(M31::from(123), M31::zero(), M31::zero(), M31::zero());
```
zmalatrax, line 183:
```suggestion
        let value: QM31 = QM31::from_m31_array([123, 0, 0, 0].map(Into::into));
```

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        memory.insert(addr, value);
+
+        let retrieved = memory.get_data(addr);
+        assert_eq!(retrieved, M31::from(123)); // Should get the projected M31 value
```
zmalatrax, line 188:
```suggestion
        assert_eq!(memory.get_data(addr), M31::from(123));
```

crates/runner/src/memory/mod.rs
zmalatrax: If you were to keep returning 0, you can do `unwrap_or_default` rather than `unwrap_or_else`

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// * `addr` - The `M31` memory address to read from.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value at the address is not a base field element.
+    pub fn get_data(&self, addr: M31) -> M31 {
+        let qm31_value = self.get_instruction(addr);
+        assert!(qm31_value.1.is_zero());
+        assert!(qm31_value.0 .1.is_zero());
+        qm31_value.0 .0
```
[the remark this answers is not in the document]
obatirou: Trunk automatic lint

zmalatrax (changes requested):
Good, just need to refactor the error handling strategy imo

crates/runner/src/memory/mod.rs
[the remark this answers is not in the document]
zmalatrax: oh ok, I didn't know, strange 🤔

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    pub fn insert_slice(&mut self, start_addr: M31, values: &[QM31]) {
+        if values.is_empty() {
+            return;
+        }
+
+        // Check that the entire slice fits within memory limits
+        let start_address = start_addr.0 as usize;
+        let slice_len = values.len();
+        // Since we already checked for empty slice, slice_len >= 1
+        let last_addr = start_addr.0.saturating_add((slice_len - 1) as u32);
```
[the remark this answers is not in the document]
zmalatrax: Ok! I prefer to avoid relying on another (subsequent) function to catch an error which can already be handled

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+impl Memory {
+    /// Checks if a given memory address is within the allowed range (`0` to `2^MAX_MEMORY_SIZE_BITS`).
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `M31` address to validate.
+    ///
+    /// # Errors
+    ///
+    /// Returns `MemoryError::AddressOutOfBounds` if the address exceeds the maximum allowed size.
```
zmalatrax, line 39: You can wrap `MemoryError::AddressOutOfBounds` in square brackets for it to be referenced (and access the definition through Ctrl+Click)

```suggestion
    /// Returns [`MemoryError::AddressOutOfBounds`] if the address exceeds the maximum allowed size.
```

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// be simple field elements. Returns an error if the retrieved `QM31` value cannot
+    /// be projected to the base field (i.e., its extension components are non-zero).
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `M31` memory address to read from.
+    ///
+    /// # Errors
+    ///
+    /// Returns `MemoryError::BaseFieldProjectionFailed` if the value at the address
```
zmalatrax, line 77: wrap the MemoryError in square brackets

crates/runner/src/memory/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+        let start_address = start_addr.0 as usize;
+        let slice_len = values.len();
+        // Since we already checked for empty slice, slice_len >= 1
+        let last_addr = start_addr.0.checked_add((slice_len - 1) as u32).ok_or(
+            MemoryError::AddressOutOfBounds {
+                address: start_addr,
+                max_address: 1 << MAX_MEMORY_SIZE_BITS,
+            },
+        )?;
+        Self::validate_address(M31::from(last_addr))?;
```
zmalatrax, line 146:
```suggestion
        Self::validate_address(last_addr.into())?;
```

zmalatrax (approved):
lgtm

PR #9, opened 2025-06-09
feat(compiler): add error management in lexer

enitrat:


crates/compiler/parser/src/lexer.rs
```diff
@@ -2,20 +2,116 @@
...
+pub enum NumberParseError {
+    /// The number is too large to fit in the target integer type (u32)
+    Overflow,
+    /// The number format is invalid (e.g., invalid characters for the base)
+    InvalidFormat,
+    /// An unknown parsing error occurred
+    Unknown(String),
+}
+
+impl fmt::Display for LexingError {
```
obatirou, line 34: I like thiserror. It is also used by SW in stwo. Also those errors do not implement the Error trait unless I miss something?
https://google.github.io/comprehensive-rust/error-handling/thiserror.html

PR #11, opened 2025-06-09
feat(vm): init instructions struct

zmalatrax:
Closes CORE-755

### DIFF
--- a/crates/runner/src/vm/instructions/mod.rs
+++ b/crates/runner/src/vm/instructions/mod.rs
@@ -8,7 +8,7 @@
 //! - Some instructions might use less than 3 arguments.
 //!
 //! A QM31 is made of 4 M31 (extension field of CM31, the extension field of M31).
-//! This is why we can represent instructions as a QM31.
+//! This is why instructions can be represented as a QM31.
 
 use stwo_prover::core::fields::{m31::M31, qm31::QM31};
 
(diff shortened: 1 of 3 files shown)


crates/runner/src/vm/instructions/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl From<QM31> for Instruction {
+    fn from(instruction: QM31) -> Self {
+        let [op, args @ ..] = instruction.to_m31_array();
+        Self { op, args }
+    }
+}
+
+impl<T: Into<M31>> From<[T; 4]> for Instruction {
+    fn from(instruction: [T; 4]) -> Self {
```
obatirou, line 27: Should rustdoc be added on functions?

crates/runner/src/vm/instructions/mod.rs
```diff
@@ -0,0 +1,57 @@
+use stwo_prover::core::fields::{m31::M31, qm31::QM31};
```
obatirou, line 1: Maybe add an introduction comment on how instruction are encoded explaining why QM31 is used

crates/runner/src/vm/instructions/mod.rs
zmalatrax: mmh I think that From structs are self-explanatory and don't need rustdoc 🤔

PR #21, opened 2025-06-11
feat(vm): add jump instructions

zmalatrax:
Closes CORE-750

### DIFF
--- a/crates/runner/src/vm/instructions/jump.rs
+++ b/crates/runner/src/vm/instructions/jump.rs
@@ -1,0 +1,485 @@
+//! JUMP instructions for the Cairo M VM.
+
+use crate::{
+    memory::{Memory, MemoryError},
+    vm::{instructions::Instruction, State},
+};
+
+/// Jump to the absolute address of the sum of the memory values at the offsets `fp + off0` and `fp + off1`.
+///
+/// CASM equivalent:
+/// ```casm
+/// jmp abs [fp + off0] + [fp + off1]
+/// ```
+pub fn jmp_abs_add_fp_fp(
+    memory: &mut Memory,
+    state: State,
+    instruction: Instruction,
+) -> Result<State, MemoryError> {
+    let [off0, off1, _] = instruction.args;
+    let offset = memory.get_data(state.fp + off0)? + memory.get_data(state.fp + off1)?;
+
+    Ok(state.jump_abs(offset))
+}
+
+/// Jump to the absolute address of the sum of the memory value at the offset `fp + off0` and the immediate value `imm`.
+///
+/// CASM equivalent:
+/// ```casm
+/// jmp abs [fp + off0] + imm
+/// ```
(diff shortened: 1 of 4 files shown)


crates/runner/src/vm/instructions/jump.rs
```diff
@@ -0,0 +1,485 @@
...
+    state: State,
+    instruction: Instruction,
+) -> Result<State, MemoryError> {
+    let [off0, imm, _] = instruction.args;
+    let offset = memory.get_data(state.fp + off0)? * imm;
+
+    Ok(state.jump_abs(offset))
+}
+
+/// Relative Jump - Add to PC the sum of the memory values at the offsets `fp + off0` and `fp + off1`.
```
obatirou, line 127: nit: The comments for relative jumps are inconsistent with those for absolute jumps.
```
/// Jump with a relative offset of the sum of the memory values at the offsets ...
```

crates/runner/src/vm/instructions/jump.rs
```diff
@@ -0,0 +1,485 @@
...
+}
+
+/// Relative Jump - Add to PC the immediate value `imm`.
+///
+/// CASM equivalent:
+/// ```casm
+/// jmp rel imm
+/// ```
+pub fn jmp_rel_imm(
+    _memory: &mut Memory,
```
obatirou, line 198: nit: 
```suggestion
    _: &mut Memory,
```

crates/runner/src/vm/instructions/jump.rs
zmalatrax: At first I did something like this, but I wasn't convinced ; wdyt ?
```rust
Jump to the relative address of the sum of the memory values at the offsets `fp + off0` and `fp + off1`.
```

crates/runner/src/vm/instructions/jump.rs
```diff
@@ -0,0 +1,483 @@
...
+}
+
+/// Absolute Jump - Replace PC by the immediate value `imm`.
+///
+/// CASM equivalent:
+/// ```casm
+/// jmp abs imm
+/// ```
+pub const fn jmp_abs_imm(
+    _memory: &mut Memory,
```
obatirou, line 84:
```suggestion
    _: &mut Memory,
```

crates/runner/src/vm/instructions/jump.rs
```diff
@@ -0,0 +1,483 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use num_traits::Zero;
+    use stwo_prover::core::fields::m31::M31;
+
+    use super::*;
+
+    const JMP_REL_INITIALSTATE: State = State {
```
obatirou, line 248:
```suggestion
    const JMP_REL_INITIAL_STATE: State = State {
```

obatirou (approved):

PR #31, opened 2025-06-12
test(vm): add iterative fibonacci

obatirou:
Closes CORE-774


crates/runner/src/vm/mod.rs
```diff
@@ -331,4 +331,89 @@ mod tests {
...
+            Instruction::from([3, 0, 1, 0]), // store_sub_fp_imm: [fp+0] = [fp+0] - 1 (counter--)
+            // Jump back to condition check
+            Instruction::from([15, 3, 0, 0]), // jmp_abs_imm: jmp abs 3
+        ];
+        let instructions_len = instructions.len() as u32;
+        let program = Program::from(instructions);
+        let mut vm = VM::try_from(program).unwrap();
+
+        let result = vm.execute();
+        assert!(result.is_ok());
```
zmalatrax, line 391: result variable is not necessary

```suggestion
        assert!(vm.execute().is_ok());
```

crates/runner/src/vm/mod.rs
```diff
@@ -331,4 +331,89 @@ mod tests {
...
+            M31(fib(n + 1))
+        );
+    }
+
+    #[test]
+    fn test_execute_fibonacci() {
+        let test_cases = [0, 1, 2, 3, 10, 20];
+        for &n in &test_cases {
+            run_fib_test(n);
+        }
```
zmalatrax, line 417:
```suggestion
        [0, 1, 2, 3, 10, 20].iter().for_each(|n| run_fib_test(*n));
```

Considering the simplicity of the test, both versions are idiomatic to be honest

PR #36, opened 2025-06-13
feat(vm): add call instructions

zmalatrax:
Closes CORE-749

### DIFF
--- a/crates/runner/src/vm/instructions/call.rs
+++ b/crates/runner/src/vm/instructions/call.rs
@@ -1,0 +1,286 @@
+//! CALL and RET instructions for the Cairo M VM.
+//!
+//! CALL instructions handle functions calls, creating new frames.
+//! There are relative and absolute function calls.
+//!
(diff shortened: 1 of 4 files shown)


crates/runner/src/vm/instructions/call.rs
```diff
@@ -0,0 +1,287 @@
...
+//!
+//! The function call is performed by:
+//! - Storing FP in memory at fp + off0.
+//! - Storing the return address in memory at fp + off0 + 1.
+//! - Updating FP for the new frame: fp + off0 + 2.
+//! - Updating PC to the function address based on off1.
+//!
+//! RET instructions returns control to the caller:
+//! - Restore FP from memory, stored at fp - 2.
+//! - Update PC to the return address, stored at fp - 1.
```
obatirou, line 21: Where are stored the return values?
When there is a ret, how one would access those?
Last time this was discussed, I thought it would be at the beginning/before the frame:
return values, then old fp then return pc

PR #39, opened 2025-06-13
fix(compiler): Fix Double Allocation for Aggregate Literals in MIR Generation

gael-bigot:


enitrat (commented):
lgtm, remove the TODO in mir_generation_tests.rs:69-71

PR #41, opened 2025-06-13
feat(compiler): initial MIR to CASM codegen

enitrat:


crates/compiler/codegen/tests/snapshots/codegen_tests__control_flow_if_else.snap
```diff
@@ -0,0 +1,33 @@
...
+        return 2;
+    }
+}
+============================================================
+Generated CASM:
+test_if_else:
+test_if_else:
+test_if_else_0:
+   0: 4 -4 0               // Store: [fp+0] = [fp+-4]
+   1: 3 0 1 0              // Equality check: [fp+1] = [fp+0] - 0
```
gael-bigot, line 25: Same here tests against 0 can be optimized.

crates/compiler/codegen/tests/snapshots/codegen_tests__control_flow_if_else.snap
```diff
@@ -0,0 +1,33 @@
...
+   2: 31 1 4               // if [fp+1] != 0 jmp rel test_if_else_2
+   3: 20 4                 // jump abs test_if_else_1
```
gael-bigot, line 27: Why absolute jumps vs relative? I think relative is better so that a compiled CASM function will always run correctly if we move it around in the bytecode.

crates/compiler/codegen/tests/snapshots/codegen_tests__functions_fib.snap
```diff
@@ -0,0 +1,68 @@
...
+
+============================================================
+Generated CASM:
+main:
+   0: 6 0 0                // Store immediate: [fp+0] = 0
+   1: 6 1 1                // Store immediate: [fp+1] = 1
+   2: 6 2 10               // Store immediate: [fp+2] = 10
+   3: 4 0 3                // Arg 0: [fp + 3] = [fp + 0]
+   4: 4 1 4                // Arg 1: [fp + 4] = [fp + 1]
+   5: 4 2 5                // Arg 2: [fp + 5] = [fp + 2]
```
gael-bigot, line 38: Once again this is very redundant but might be an optimization for later on.

gael-bigot (dismissed):
Otherwise lgtm

PR #47, opened 2025-06-16
feat(vm): export traces

obatirou:
Closes CORE-795

### DIFF
--- a/crates/runner/src/vm/mod.rs
+++ b/crates/runner/src/vm/mod.rs
@@ -463,10 +463,7 @@ mod tests {
         // Expected serialized data:
         // Entry 1: fp=2, pc=0.
         // Entry 2: fp=2, pc=1.
-        let expected_bytes: Vec<u8> = [2u32, 0u32, 2u32, 1u32]
-            .iter()
-            .flat_map(|&x| x.to_le_bytes())
-            .collect();
+        let expected_bytes = Vec::from([2, 0, 2, 1].map(u32::to_le_bytes).as_flattened());
 
         assert_eq!(serialized_trace, expected_bytes);
     }
@@ -584,10 +581,7 @@ mod tests {
         // Expected serialized data:
         // Entry 1: fp=2, pc=0.
         // Entry 2: fp=2, pc=1.
-        let expected_bytes: Vec<u8> = [2u32, 0u32, 2u32, 1u32]
-            .iter()
-            .flat_map(|&x| x.to_le_bytes())
-            .collect();
+        let expected_bytes = Vec::from([2, 0, 2, 1].map(u32::to_le_bytes).as_flattened());
 
         assert_eq!(file_contents, expected_bytes);
     }
--- a/crates/runner/src/memory/mod.rs
+++ b/crates/runner/src/memory/mod.rs
@@ -495,11 +495,11 @@ mod tests {
 
         // Entry 1: addr=10, value=[1, 2, 3, 4]
         // Entry 2: addr=20, value=[5, 6, 7, 8]
-        let expected_bytes: Vec<u8> =
-            [10u32, 1u32, 2u32, 3u32, 4u32, 20u32, 5u32, 6u32, 7u32, 8u32]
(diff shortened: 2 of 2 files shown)


crates/runner/src/vm/mod.rs
```diff
@@ -165,6 +170,52 @@ impl VM {
             .flat_map(u32::to_le_bytes)
             .collect()
     }
+
+    /// Writes the serialized trace to a binary file.
+    ///
+    /// This function serializes the trace using [`serialize_trace()`](Self::serialize_trace)
+    /// and writes the resulting bytes to the specified file path.
```
zmalatrax, line 177: There is a description of the encoded format for `write_binary_memory_trace` but not for this function

crates/runner/src/vm/mod.rs
```diff
@@ -498,4 +552,90 @@ mod tests {
...
+        assert!(result.is_ok());
+
+        // Read the file back and verify its contents.
+        let mut file = File::open(temp_file_path).unwrap();
+        let mut file_contents = Vec::new();
+        file.read_to_end(&mut file_contents).unwrap();
+
+        // Compare with the expected serialized trace.
+        let expected_trace = vm.serialize_trace();
+        assert_eq!(file_contents, expected_trace);
```
zmalatrax, line 581:
```suggestion
        assert_eq!(file_contents, vm.serialize_trace());

File: crates/runner/src/vm/mod.rs [hunk 3]
```diff
@@ -498,4 +552,90 @@ mod tests {
...
+        assert!(result.is_ok());
+
+        // Read the file back and verify its contents.
+        let mut file = File::open(temp_file_path).unwrap();
+        let mut file_contents = Vec::new();
+        file.read_to_end(&mut file_contents).unwrap();
+
+        // Compare with the expected serialized memory trace.
+        let expected_memory_trace = vm.memory.serialize_trace();
+        assert_eq!(file_contents, expected_memory_trace);
```
zmalatrax, line 624:
```suggestion
        assert_eq!(file_contents, vm.memory.serialize_trace());

File: crates/runner/src/vm/mod.rs [hunk 4]
```diff
@@ -498,4 +552,90 @@ mod tests {
...
+        ];
+        let program = Program::from(instructions);
+        let mut vm = VM::try_from(program).unwrap();
+
+        // Execute the program to generate a trace.
+        assert!(vm.execute().is_ok());
+
+        // Write the trace to a temporary file.
+        let temp_file_path = "test_trace.bin";
+        let result = vm.write_binary_trace(temp_file_path);
```
zmalatrax, line 571: if the test fails prior to deleting this file, it won't be deleted?

Not really a "temporary file" in the sense that it is not in the temp directory etc. ; but I understood the purpose

--> Should we consider introducing temporary files creation (e.g. with `tempfile` rust crate) ?

crates/runner/src/vm/mod.rs
obatirou: I thought it would be duplicating the rustdoc of serialize_trace, I think that is ok to let as is?

crates/runner/src/vm/mod.rs
```diff
@@ -463,11 +463,10 @@ mod tests {
         // Expected serialized data:
         // Entry 1: fp=2, pc=0.
         // Entry 2: fp=2, pc=1.
-        let mut expected_bytes = Vec::new();
-        expected_bytes.extend_from_slice(&2u32.to_le_bytes());
-        expected_bytes.extend_from_slice(&0u32.to_le_bytes());
-        expected_bytes.extend_from_slice(&2u32.to_le_bytes());
-        expected_bytes.extend_from_slice(&1u32.to_le_bytes());
+        let expected_bytes: Vec<u8> = [2u32, 0u32, 2u32, 1u32]
```
zmalatrax, line 466: See previous suggestion, apply if used

zmalatrax (approved):

PR #60, opened 2025-06-17
feat(vm): CLI to load and run program from artifact

enitrat:


crates/runner/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+        })
+        .collect();
+
+    // Create the program and VM
+    let program = Program { instructions };
+    let mut vm = VM::try_from(program)?;
+
+    // Run from the entry point
+    let fp_offset = 10;
+    vm.run_from_entrypoint(pc_entrypoint as u32, fp_offset)?;
```
obatirou, line 80: Why does fp is hardcoded but not pc?

crates/runner/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+            }
+        })
+        .collect();
+
+    let program = Program { instructions };
+    let mut vm = VM::try_from(program)?;
+
+    // TODO: Get entrypoint information from the compiled program to know how many args / return data to allocate.
+    const FP_OFFSET: u32 = 3;
+    vm.run_from_entrypoint(pc_entrypoint as u32, FP_OFFSET)?;
```
zmalatrax, line 67: currently hardcoded `FP_OFFSET` translates as executing an entrypoint with "1 argument or return value, old fp, return address", right?

zmalatrax (approved):
lgtm (lfg)

PR #66, opened 2025-06-17
tests(vm-compiler): diff testing for .cm programs

enitrat:
Close CORE-871

### DIFF
--- a/.github/workflows/trunk-check.yml
+++ b/.github/workflows/trunk-check.yml
@@ -23,7 +23,9 @@ jobs:
           submodules: recursive
 
       - name: Set up Rust toolchain
-        run: rustup show
+        run: |
+          rustup show
+          rustup component add clippy
 
       - name: Rust cache
         uses: Swatinem/rust-cache@v2


crates/compiler/src/lib.rs
```diff
@@ -1,7 +1,157 @@
...
+    db: &CompilerDatabase,
+    source: SourceProgram,
+    _options: CompilerOptions,
+) -> Result<CompilerOutput> {
+    // Parse the program
+    let parsed_program = parse_program(db, source);
+
+    if !parsed_program.diagnostics.is_empty() {
+        return Err(CompilerError::ParseErrors(parsed_program.diagnostics));
+    }
```
obatirou, line 97: warnings are part of diagnostics no?
shouldn't there is a preprocessing phase to differentiate the 2?
Here if there is only warnings it will send back an error, it is the desired behavior?

crates/runner/src/lib.rs
```diff
@@ -1,2 +1,83 @@
...
+) -> Result<RunnerOutput> {
+    let pc_entrypoint = program.get_entry_point(entry_point).ok_or_else(|| {
+        RunnerError::EntryPointNotFound(
+            entry_point.to_string(),
+            program.entry_points.keys().cloned().collect(),
+        )
+    })?;
+
+    let instructions: Vec<Instruction> =
+        program.instructions.iter().map(Instruction::from).collect();
```
zmalatrax, line 63:
```suggestion
        program.instructions.iter().map(Into::into).collect();
```

crates/compiler/src/lib.rs
```diff
@@ -1,7 +1,157 @@
...
+impl std::error::Error for CompilerError {}
+
+/// Options for compilation
+#[derive(Debug, Clone, Default)]
+pub struct CompilerOptions {
+    /// Enable verbose output
+    pub verbose: bool,
+}
+
+/// Compilation output including the compiled program and any diagnostics
```
obatirou, line 53: do not match field as only warnings are part of the output

crates/compiler/src/lib.rs
```diff
@@ -1,7 +1,157 @@
...
+}
+
+/// Compiles a Cairo-M program from a SourceProgram
+///
+/// This is a lower-level API that allows reusing a database instance
+/// for incremental compilation scenarios.
+pub fn compile_from_source(
+    db: &CompilerDatabase,
+    source: SourceProgram,
+    _options: CompilerOptions,
```
zmalatrax, line 90: Previously implemented `verbose` flag behavior is not replicated here

crates/runner/src/lib.rs
```diff
@@ -1,2 +1,83 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum RunnerError {
+    #[error("Entry point '{0}' not found. Available entry points: {1:?}")]
+    EntryPointNotFound(String, Vec<String>),
+
+    #[error("VM error: {0}")]
+    VmError(#[from] VmError),
+
+    #[error("Failed to read return value: {0}")]
+    ReturnValueError(String),
```
obatirou, line 22:
```suggestion
     ReturnValueError(#[from] MemoryError),
```
If not mistaken, the error type is know as it is a get on memory

crates/runner/tests/diff_tests.rs
```diff
@@ -0,0 +1,261 @@
...
+
+diff_test!(
+    test_factorial,
+    "factorial.cm",
+    "main",
+    rust_factorial,
+    "Factorial computation for n=10"
+);
+
+// 4. Power (recursive)
```
zmalatrax, line 112: Why naive implementation rather than exponentiation by squaring?

crates/runner/tests/diff_tests.rs
```diff
@@ -0,0 +1,261 @@
...
+
+diff_test!(
+    test_power,
+    "power.cm",
+    "main",
+    rust_power,
+    "3^10 computation"
+);
+
+// 5. Sum of first N numbers
```
zmalatrax, line 132: specify that it is recursive ? As in 4. power

crates/compiler/src/lib.rs
```diff
@@ -1,7 +1,157 @@
...
+    if !semantic_errors.is_empty() {
+        return Err(CompilerError::SemanticErrors(semantic_errors));
+    }
+
+    // Collect warnings
+    let warnings: Vec<_> = semantic_diagnostics
+        .iter()
+        .filter(|d| d.severity == DiagnosticSeverity::Warning)
+        .cloned()
+        .collect();
```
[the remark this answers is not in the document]
zmalatrax: It implies having mutable objects

Here is another suggestion, the filtering on Warning and Errors is required due to `DiagnosticSeverity` having 2 other variants Hints and Info
```suggestion
    let (semantic_errors, warnings): (Vec<_>, Vec<_>) = semantic_diagnostics
        .into_iter()
        .filter(|d| {
            matches!(
                d.severity,
                DiagnosticSeverity::Error | DiagnosticSeverity::Warning
            )
        })
        .partition(|d| d.severity == DiagnosticSeverity::Error);

    if !semantic_errors.is_empty() {
        return Err(CompilerError::SemanticErrors(semantic_errors));
    }
```

crates/compiler/src/lib.rs
obatirou: Oh ok, I was confused by the severity
What is the purpose of the severity of a diagnostics?
```rust
pub enum DiagnosticSeverity {
    Error,
    Warning,
    Info,
    Hint,
}

```

crates/compiler/src/main.rs
```diff
@@ -27,76 +23,83 @@ struct Args {
...
-                eprintln!("Failed to generate program: {e}");
-                std::process::exit(1);
+    // Read the input file
+    let source_text = match fs::read_to_string(&args.input) {
+        Ok(content) => content,
+        Err(e) => {
+            eprintln!("Error reading file '{}': {}", args.input.display(), e);
+            process::exit(1);
+        }
+    };
```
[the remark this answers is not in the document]
enitrat: ✅ Fixed\! Changed from match expression to `unwrap_or_else`:
```rust
let source_text = fs::read_to_string(&args.input).unwrap_or_else(|e| {
    eprintln\!("Error reading file '{}': {}", args.input.display(), e);
    process::exit(1);
});
```

crates/compiler/src/main.rs
```diff
@@ -27,76 +23,83 @@ struct Args {
...
-                for (name, pc) in &program.entry_points {
-                    println!("    {name}: PC {pc}");
+            // Serialize the program to JSON
+            let json = match sonic_rs::to_string_pretty(&*output.program) {
+                Ok(json) => json,
+                Err(e) => {
+                    eprintln!("Failed to serialize program: {}", e);
+                    process::exit(1);
+                }
+            };
```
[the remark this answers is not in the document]
enitrat: ✅ Fixed\! Updated to use `unwrap_or_else`:
```rust
let json = sonic_rs::to_string_pretty(&*output.program).unwrap_or_else(|e| {
    eprintln\!("Failed to serialize program: {}", e);
    process::exit(1);
});
```

crates/compiler/src/lib.rs
enitrat: The current implementation is correct. The parser only generates errors (not warnings), so returning an error for any diagnostics at line 95-97 is appropriate. The semantic phase (lines 102-117) properly separates errors from warnings:
- Parse errors: Always fail compilation
- Semantic errors: Fail compilation
- Semantic warnings: Collected and returned with the compiled output

No changes needed.

crates/runner/src/lib.rs
```diff
@@ -1,2 +1,83 @@
...
+pub struct RunnerOptions {
+    // Empty for now
+}
+
+/// Result of running a Cairo program
+#[derive(Debug)]
+pub struct RunnerOutput {
+    /// The return value of the program
+    pub return_value: u32,
+    /// The final VM state (for advanced usage)
```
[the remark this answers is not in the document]
enitrat: ✅ Fixed\! Clarified the comment to be more explicit:
```rust
/// The final VM state including memory and registers (for advanced usage)
pub vm: VM,
```

crates/compiler/src/lib.rs
```diff
@@ -1,7 +1,157 @@
...
+    pub verbose: bool,
+}
+
+/// Compilation output including the compiled program and any diagnostics
+#[derive(Debug)]
+pub struct CompilerOutput {
+    /// The compiled program
+    pub program: Arc<CompiledProgram>,
+    /// Any warnings generated during compilation
+    pub warnings: Vec<Diagnostic>,
```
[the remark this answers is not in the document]
enitrat: The CompilerOutput struct only contains warnings because errors prevent compilation from succeeding. The compile function returns Result<CompilerOutput>, so:
- If there are errors (parse or semantic), compilation fails and returns Err(CompilerError)
- If compilation succeeds, it returns Ok(CompilerOutput) with the compiled program and any warnings
This is a common pattern where errors are fatal but warnings are informational.

crates/compiler/src/lib.rs
enitrat: Good catch\! The verbose flag is currently passed through the CompilerOptions but not used within the compile_from_source function. This would need to be implemented to show MIR output when verbose is enabled. The implementation was intentionally deferred to keep the PR focused on the API structure.

crates/compiler/src/lib.rs
```diff
@@ -1,7 +1,157 @@
...
+    ParseErrors(Vec<Diagnostic>),
+    /// Semantic validation errors occurred
+    SemanticErrors(Vec<Diagnostic>),
+    /// MIR generation failed
+    MirGenerationFailed,
+    /// Code generation failed
+    CodeGenerationFailed(String),
+}
+
+impl std::fmt::Display for CompilerError {
```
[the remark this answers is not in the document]
enitrat: The RunnerError in runner/src/lib.rs does use thiserror. For CompilerError in compiler/src/lib.rs, the manual implementation was chosen because:
1. It provides a simple, consistent error type across the compiler crate
2. The Display messages are straightforward and don't need the macro overhead
3. It avoids adding another dependency to the core compiler crate

Both approaches are valid - thiserror is great for more complex error types with source chains.

crates/compiler/src/lib.rs
enitrat: Good point\! The comment should be updated to:
```rust
/// Compilation output including the compiled program and any warnings
```
This accurately reflects that the struct only contains warnings, as errors cause compilation to fail entirely.

crates/compiler/src/lib.rs
enitrat: Error => fails
Warning => Draw attention, but still compiles (unused var, etc), could be sub-optimal code
Info | Hints => informational, tbd (for now no such diag emitted)

crates/runner/tests/diff_tests.rs
enitrat: The naive implementation was chosen to match the simplicity of the corresponding Cairo-M test program. The goal is to ensure the VM produces identical results to the Rust reference implementation, not to optimize the reference implementation itself. For testing purposes, a simple, clearly correct implementation is preferable.

crates/runner/CLAUDE.md
```diff
@@ -92,3 +92,67 @@ The runner integrates with:
...
 - `stwo-prover`: Generates traces compatible with Stwo proving system
 - Binary trace format for efficient proof generation
+
+### Coding Guidelines
+
+#### Error Handling
+
+- Use `thiserror` for custom error types to avoid boilerplate in `Display` and
+  `Error` implementations
+- Use `anyhow` for error handling in binaries with `with_context()` for
```
zmalatrax, line 102: Note that `anyhow` is not currently used in the binaries

enitrat: yeah, would be nice to though

PR #67, opened 2025-06-17
feat(prover): implement adapter

AntoineFONDEUR:
Closes CORE-817


crates/prover/src/adapter/io.rs
ClementWalter, line 1: - let's use a `BufReader` to reduce IO while making sure the memory usage remains limited
- don't collect the iterator, just use them directly in the `adapt_from_iter`

crates/prover/src/adapter/io.rs
```diff
@@ -0,0 +1,70 @@
...
+#[repr(C)]
+#[derive(Copy, Clone, Default, Pod, Zeroable)]
+pub struct TraceEntry {
+    pub pc: u32,
+    pub fp: u32,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, Default, Pod, Zeroable, Debug)]
+pub struct MemEntry {
```
ClementWalter, line 26: don't use abbreviation, so `MemoryEntry`

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,326 @@
...
+            fp: entry.fp,
+        }
+    }
+}
+
+//TODO: add prev value
+#[derive(Debug, Default)]
+pub struct StateData {
+    pub state: VmState,
+    pub memory_args: [(u32, QM31, QM31, u32, u32); 4], // (addr, val, prev_val, prev_clock, clock)
```
ClementWalter, line 25: define a struct for MemoryArg

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,326 @@
...
+        memory_cache: &mut MemoryCache,
+    ) where
+        I: Iterator<Item = MemEntry>,
+    {
+        let mut state_data = StateData {
+            state,
+            memory_args: Default::default(),
+        };
+
+        let instruction = memory.next().unwrap();
```
ClementWalter, line 86: no `unwrap`

crates/prover/src/adapter/io.rs
ClementWalter: and `address` and `value`

crates/prover/src/adapter/io.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub addr: u32,
+    pub val: [u32; 4],
```
ClementWalter, line 28: create a `MemoryValue` type

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,326 @@
...
+            pc: entry.pc,
+            fp: entry.fp,
+        }
+    }
+}
+
+//TODO: add prev value
+#[derive(Debug, Default)]
+pub struct StateData {
+    pub state: VmState,
```
ClementWalter, line 24: use `registers` and not `state`

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+impl MemoryCache {
+    pub fn new() -> Self {
+        Self {
+            clock_cache: HashMap::new(),
+            initial_memory: HashMap::new(),
+            current_memory: HashMap::new(),
+        }
+    }
```
ClementWalter, line 27: this is `default` you derived from if I'm not mistaken

crates/prover/src/adapter/io.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut entry = IoTraceEntry::default();
+        self.0
+            .read_exact(bytes_of_mut(&mut entry))
+            .ok()
+            .map(|_| entry.into())
+    }
+}
+
+// Standalone function to create a TraceIter from a file path
+pub fn trace_iter_from_path(path: &Path) -> Result<TraceFileIter, VmImportError> {
```
Eikix, line 79: I think this can also be a TryFrom trait

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,361 @@
+use std::collections::HashMap;
+
+use stwo_prover::core::fields::qm31::QM31;
+
+use crate::adapter::instructions::MemoryArg;
+
+#[derive(Copy, Clone, Default, Debug)]
+pub struct MemoryEntry {
```
Eikix, line 8: what is the difference between this memoryentry and io::MemoryEntry?

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,723 @@
...
+    ) -> Result<(), InstructionError>
+    where
+        I: Iterator<Item = MemoryEntry>,
+    {
+        let mut state_data = StateData {
+            registers,
+            memory_args: Default::default(),
+        };
+
+        let instruction = memory
```
Eikix, line 103: how do you know that the instruction is located at the next value in memory?
i thought the instruction depended on memory[pc]?

Eikix: apart from that I think it'd be nice to have a Opcode struct that has a From<QM31>, such that you can jsut do instruction.value.into().opcode_id()

crates/prover/src/adapter/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+                initial_memory: memory.clone(),
+                final_memory: memory,
+            },
+            instructions,
+        };
+    };
+
+    instructions.initial_registers = first.into();
+    instructions
+        .push_instr(&mut memory, first.into(), clock, &mut memory_cache)
```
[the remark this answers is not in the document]
AntoineFONDEUR: Maybe this also answers this question : 

> how do you know that the instruction is located at the next value in memory?
> i thought the instruction depended on memory[pc]?

The VM outputs a log that writes every memory reads/writes chronologically. So if the VM reads `[addr, instruction]` and that the opcode is for instance `store_imm`, the runner will write at [fp+instruction_off2] the value instruction_imm. The result of this opcode execution in the log is adding [addr, instruction] and then [fp+instruction_off2]. So by knowing the number of RW of each opcode, we can just go through the memory and trace sequentially, re-running the VM in a way.

crates/prover/src/adapter/memory.rs
AntoineFONDEUR: Clément says:

> don't use the MemoryEntry from the io adapter, so that if the file format change in some way, you don't need to update the memory struct elsewhere (as you did for the trace entry actually)

crates/prover/src/adapter/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    mem_iter: I,
+    trace_iter: J,
+) -> ProverInput {
+    let mut instructions = Instructions::default();
+    let mut memory = mem_iter.into_iter();
+    let mut trace = trace_iter.into_iter();
+    let mut clock = 1;
+    let mut memory_cache = MemoryCache::default();
+
+    let Some(first) = trace.next() else {
```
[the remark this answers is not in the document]
AntoineFONDEUR: I'll throw an error here 👍

crates/prover/src/adapter/memory.rs
Eikix: maybe the io file should only have IoMemoryEntry, and you keep this MemoryEntry and you just implem From trait once

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use stwo_prover::core::fields::m31::M31;
+use stwo_prover::core::fields::qm31::QM31;
+
+use crate::adapter::instructions::MemoryArg;
+
+#[derive(Copy, Clone, Default, Debug)]
+pub struct MemoryEntry {
+    pub address: u32,
+    pub value: [u32; 4],
```
ClementWalter, line 11: add `clock` also

ClementWalter, line 11: use `M31` types (see what you did in MemoryBoundaries: `(M31, QM31, M31)`)

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub fn push(&mut self, mem_entry: MemoryEntry, clock: u32) -> MemoryArg {
+        let prev_clock = self
+            .clock_cache
+            .get(&mem_entry.address)
+            .copied()
+            .unwrap_or_else(|| {
+                // If the memory cell is uninitialized, mark this address as an initial one with dummy value and clock
+                // and return previous clock as zero.
+                self.initial_memory
+                    .insert(mem_entry.address, Default::default());
```
ClementWalter, line 61: why not pushing the actual value and clock?

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+        let old_value = self
+            .current_memory
+            .insert(mem_entry.address, (mem_entry.value, clock))
+            .unwrap_or_default();
+
+        // Like so, components can systematically consume the previous (addr, val, clock) and produce the new (addr, val, clock).
+        // And the memory component can produce the initial (addr, default_val, default_clock) and consume the final (addr, val, clock).
+        MemoryArg {
+            address: mem_entry.address.into(),
+            prev_val: QM31::from_u32_unchecked(
```
ClementWalter, line 77:
```suggestion
            prev_value: QM31::from_u32_unchecked(
```

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+
+#[derive(Debug, Default, Clone)]
+pub struct MemoryBoundaries {
+    pub initial_memory: Vec<(M31, QM31, M31)>,
+    pub final_memory: Vec<(M31, QM31, M31)>,
+}
+
+#[derive(Debug, Default)]
+pub struct MemoryCache {
+    clock_cache: HashMap<u32, u32>,
```
ClementWalter, line 46: I think that you don't need the clock cache, just insert into the current_memory to get the previous value. if it's none, then append also to the `initial_memory`

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+                0
+            });
+        // No matter what (initialized cell or not), update the clock cache at the given address with the current clock.
+        self.clock_cache.insert(mem_entry.address, clock);
+        // Update the current memory at the given address with the current value and clock.
+        // If the memory cell is uninitialized, the old value is the dummy value.
+        let old_value = self
+            .current_memory
+            .insert(mem_entry.address, (mem_entry.value, clock))
+            .unwrap_or_default();
```
ClementWalter, line 71: here, just do "if none, then insert memory_entry into initial_memory"

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+pub struct MemoryBoundaries {
+    pub initial_memory: Vec<(M31, QM31, M31)>,
+    pub final_memory: Vec<(M31, QM31, M31)>,
+}
+
+#[derive(Debug, Default)]
+pub struct MemoryCache {
+    clock_cache: HashMap<u32, u32>,
+    initial_memory: HashMap<u32, ([u32; 4], u32)>,
+    current_memory: HashMap<u32, ([u32; 4], u32)>,
```
ClementWalter, line 48: can we use `M31 / QM31` here directly?

crates/prover/src/adapter/memory.rs
```diff
@@ -0,0 +1,213 @@
...
+                mem_entry.value[1],
+                mem_entry.value[2],
+                mem_entry.value[3],
+            ),
+            prev_clock: prev_clock.into(),
+            clock: clock.into(),
+        }
+    }
+
+    pub fn get_memory_boundaries(&self) -> MemoryBoundaries {
```
ClementWalter, line 94: make this an `Into`

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,474 @@
...
+        Self { id: value[0] }
+    }
+}
+
+impl From<&[u32; 4]> for Opcode {
+    /// Extract opcode from reference to instruction value array
+    fn from(value: &[u32; 4]) -> Self {
+        Self { id: value[0] }
+    }
+}
```
ClementWalter, line 41: use `Borrow` and keep only one implementation

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,474 @@
...
+pub struct Opcode {
+    id: u32,
+}
+
+impl Opcode {
+    /// Get the opcode ID as a u32
+    pub const fn id(self) -> u32 {
+        self.id
+    }
+}
```
ClementWalter, line 26: I think that either we build a full opcode table with all the details (like number of consumed memory slots) we can just drop this for now

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,474 @@
...
+
+#[derive(Debug, Default)]
+pub struct StateData {
+    pub registers: VmRegisters,
+    pub memory_args: [MemoryArg; 4],
+}
+
+#[derive(Debug, Default)]
+pub struct StatesByOpcodes {
```
ClementWalter, line 74: with the `Opcode` being an `enum` one could just have

```
// Dynamic mapping of Opcode to Vec<StateData>
pub struct StatesByOpcodes {
    map: HashMap<Opcode, Vec<StateData>>,
}
```

crates/prover/src/adapter/instructions.rs
```diff
@@ -0,0 +1,474 @@
...
+            memory_args: Default::default(),
+        };
+
+        let instruction = memory
+            .next()
+            .ok_or(InstructionError::UnexpectedEndOfMemory)?;
+        let opcode_id = Opcode::from(instruction.value).id();
+        state_data.memory_args[0] = memory_cache.push(instruction, clock);
+
+        match opcode_id {
```
ClementWalter, line 139: with the `opcode` enum implemented as above, you can remove the whole match and just gets the number of items to "next" from the memory like `memory.take(n).collect()`

crates/prover/src/adapter/instructions.rs
Eikix: I agree with the spirit of this idea

crates/prover/src/adapter/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    for trace_entry in trace_iter {
+        let mut memory_args: [MemoryArg; 4] = Default::default();
+
+        let mut opcode_entry: MemoryEntry =
+            memory_iter.next().ok_or(VmImportError::EmptyTrace)?.into();
+        opcode_entry.clock = clock.into();
+        memory_args[0] = memory.push(opcode_entry);
+
+        let opcode: Opcode = Opcode::try_from(opcode_entry.value.to_m31_array()[0].0)?;
```
ClementWalter, line 77: implement a `Opcode from QM31` and check the value is not oob

crates/common/src/opcode.rs
```diff
@@ -80,6 +87,16 @@ impl TryFrom<M31> for Opcode {
     }
 }
 
+impl TryFrom<QM31> for Opcode {
+    type Error = InstructionError;
+
+    fn try_from(value: QM31) -> Result<Self, Self::Error> {
+        let opcode_u32 = value.to_m31_array()[0].0;
```
ClementWalter, line 94: you need to assert that all other values are 0. You can just use the fact that `impl TryInto<M31> for QM31 {` exists, so `value.try_into()?.into()` should be ok

crates/prover/src/adapter/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+pub fn import_from_vm_output(
+    trace_path: &Path,
+    mem_path: &Path,
+) -> Result<ProverInput, VmImportError> {
+    let _span = span!(Level::INFO, "import_from_vm_output").entered();
+
+    let mut trace_iter = TraceFileIter::try_from(trace_path)?.peekable();
+    let mut memory_iter = MemoryEntryFileIter::try_from(mem_path)?;
+    let _program_length = memory_iter.program_length();
```
ClementWalter, line 57:
```suggestion
```

PR #69, opened 2025-06-17
chore(ci): add runner benches to CI

enitrat:
Close CORE-877

### DIFF
--- a/.trunk/configs/.yamllint.yaml
+++ b/.trunk/configs/.yamllint.yaml
@@ -1,7 +1,6 @@
 rules:
   quoted-strings:
-    required: only-when-needed
-    extra-allowed: ["{|}"]
+    required: false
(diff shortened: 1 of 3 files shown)


.trunk/configs/.yamllint.yaml
```diff
@@ -1,7 +1,6 @@
 rules:
   quoted-strings:
-    required: only-when-needed
-    extra-allowed: ["{|}"]
+    required: false
```
zmalatrax, line 3: Where did you have issues with quoted strings lint ?

enitrat: in the yaml files for workflows, I hate having un-quoted strings, I think this is a dumb lint

obatirou (approved):

PR #70, opened 2025-06-18
feat(compiler): Added optimization passes to casm generation

gael-bigot:
after building a block we apply these two passes:
- convert any useless operations (eg sub 0, mul 1) to storeDerefFp
- apply copy propagation to remove useless copies

This partially solves CORE-855 (see /crates/compiler/codegen/tests/snapshots/codegen_tests__functions_fib.snap)

TODO :
- fix fp offsets after the passes (tricky but would save space on the stack)
- fix comments on new instructions

### DIFF
--- a/crates/compiler/codegen/src/optimizations.rs
+++ b/crates/compiler/codegen/src/optimizations.rs
@@ -2,9 +2,10 @@
 //!
 //! This module contains optimization passes for CASM instructions.
 
+use std::collections::{HashMap, HashSet};
+
 use crate::builder::SymbolicInstruction;
 use crate::{CasmBuilder, CasmInstruction, CodegenResult, Opcode};
-use std::collections::{HashMap, HashSet};
 
 /// Trait for CASM optimization passes
 pub trait CasmOptimization {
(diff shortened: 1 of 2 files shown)


crates/compiler/codegen/src/optimizations.rs
```diff
@@ -0,0 +1,472 @@
...
+
+impl Default for OptimizationPipeline {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Neutral operations pass
+///
+/// This pass eliminates neutral operations (add 0, sub 0, mul 1, div 1)
```
enitrat, line 52: these operations should very likely be eliminated at the MIR level, as they are not related to CASM at all and are just identity operations - so it would be way easier to detect them in MIR.

enitrat (changes requested):
Currently we only want to focus on very targeted and simple optimizations:

1. Avoid identity operations (X + 0, X * 1, X / 1, X - 0)
2. Optimize general-purpose instructions to match CASM's specific opcodes: comparison with 0 in jumps, 
3. Avoid storing the returned value in intermediate registers

For 1) this should be done at the MIR level
For 2) and 3) these two can be 30 LOC changes, local to the functions for generating instructions.

It's too early to introduce a generic abstraction - especially considering that we do not compile all the opcodes supported by the VM and we'll likely want to, in the future, modify the generated CASM with more specific opcodes, which means that the optimizer abstraction might be less justified

PR #73, opened 2025-06-18
feat(prover): implement memory component

AntoineFONDEUR:
Closes CORE-878

### DIFF
--- a/crates/prover/src/components/memory.rs
+++ b/crates/prover/src/components/memory.rs
@@ -65,10 +65,8 @@ impl Claim {
         )
         .ilog2();
 
-        let mut packed_inputs: Vec<[PackedM31; N_M31_IN_MEMORY_ENTRY]> = Vec::new();
-
         // Pack memory entries from the prover input
-        inputs
+        let packed_inputs: Vec<[PackedM31; N_M31_IN_MEMORY_ENTRY]> = inputs
             .initial_memory
             .drain()
             .chain(inputs.final_memory.drain())
@@ -93,11 +91,10 @@ impl Claim {
             .chain(std::iter::repeat([M31::zero(); N_M31_IN_MEMORY_ENTRY]))
             .take(1 << self.log_size)
             .array_chunks::<N_LANES>()
-            .for_each(|chunk| {
-                packed_inputs.push(std::array::from_fn(|x| {
-                    PackedM31::from_array(std::array::from_fn(|y| chunk[y][x]))
-                }));
-            });
+            .map(|chunk| {
+                std::array::from_fn(|x| PackedM31::from_array(std::array::from_fn(|y| chunk[y][x])))
+            })
+            .collect();
 
         // Generate lookup data and fill the trace
         let (mut trace, mut lookup_data) = unsafe {
--- a/crates/prover/src/components/memory.rs
+++ b/crates/prover/src/components/memory.rs
@@ -1,4 +1,7 @@
-use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use num_traits::Zero;
+use rayon::iter::{
+    IndexedParallelIterator, IntoParallelIterator, IntoParallelRefMutIterator, ParallelIterator,
+};
 use stwo_air_utils::trace::component_trace::ComponentTrace;
 use stwo_air_utils_derive::{IterMut, ParIterMut, Uninitialized};
 use stwo_prover::constraint_framework::logup::LogupTraceGenerator;
@@ -8,7 +11,7 @@ use stwo_prover::constraint_framework::{
 use stwo_prover::core::backend::simd::m31::{PackedM31, LOG_N_LANES, N_LANES};
 use stwo_prover::core::backend::simd::qm31::PackedQM31;
 use stwo_prover::core::backend::simd::SimdBackend;
-use stwo_prover::core::backend::{BackendForChannel, Column};
+use stwo_prover::core::backend::BackendForChannel;
 use stwo_prover::core::channel::{Channel, MerkleChannel};
 use stwo_prover::core::fields::m31::{BaseField, M31};
 use stwo_prover::core::fields::qm31::SecureField;
@@ -20,7 +23,7 @@ use stwo_prover::core::poly::BitReversedOrder;
 use crate::adapter::memory::Memory;
 use crate::relations;
 
-const N_M31_IN_MEMORY_ENTRY: usize = 7;
+const N_M31_IN_MEMORY_ENTRY: usize = 7; // Address, value, clock, multiplicity
 
 #[derive(Clone, Default)]
 pub struct Claim {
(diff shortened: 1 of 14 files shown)


crates/prover/src/components/memory.rs
```diff
@@ -19,10 +17,14 @@ use stwo_prover::core::pcs::TreeVec;
 use stwo_prover::core::poly::circle::{CanonicCoset, CircleEvaluation};
 use stwo_prover::core::poly::BitReversedOrder;
 
+use crate::adapter::memory::MemoryBoundaries;
```
ClementWalter, line 20: the only object that should be imported from the `adapter` is the ProverInput to keep the interface as small as possible

crates/prover/src/components/memory.rs
```diff
@@ -19,10 +17,14 @@ use stwo_prover::core::pcs::TreeVec;
...
 use crate::relations;
 
-#[derive(Copy, Clone, Default)]
+const N_M31_IN_MEMORY_ENTRY: usize = 7;
+
+#[derive(Clone, Default)]
 pub struct Claim {
+    pub inputs: MemoryBoundaries,
```
ClementWalter, line 27: The `Claim` is included in the proof, but we don't want to include the whole bounds in the public memory (so far)

crates/prover/src/prover.rs
```diff
@@ -10,13 +10,16 @@ use stwo_prover::core::proof_of_work::GrindOps;
 use stwo_prover::core::prover::prove;
 use tracing::{info, span, Level};
 
+use crate::adapter::ProverInput;
 use crate::components::{Claim, Components, InteractionClaim, Relations};
 use crate::errors::ProvingError;
 use crate::preprocessed::PreProcessedTrace;
 use crate::{relations, Proof};
 
+pub(crate) const LOG_MAX_ROWS: u32 = 26;
```
ClementWalter, line 19: benchmarks showed that we want to have max 21 max rows

ClementWalter (changes requested):
- review the memory component to have 7 columns only (6 + mult)
- use par_iter_mut (see removed components) to write the trace
- remove the inputs from the claim and update `write_trace` to take the input as arg and return `Self` (like `InteractionClaim`)

crates/prover/src/components/memory.rs
```diff
@@ -19,10 +17,14 @@ use stwo_prover::core::pcs::TreeVec;
 use stwo_prover::core::poly::circle::{CanonicCoset, CircleEvaluation};
 use stwo_prover::core::poly::BitReversedOrder;
 
+use crate::adapter::memory::MemoryBoundaries;
 use crate::relations;
 
-#[derive(Copy, Clone, Default)]
+const N_M31_IN_MEMORY_ENTRY: usize = 7;
```
Eikix, line 23: right now the struct MemoryEntry has only 5 m31 afaik
maybe we mean N_M31_IN_MEMORY_PER_ROW?

there is something i dont understand: `addr, value (4), prev_value, new_value, clk, prev_clk` is 9 values, what did i miss?

crates/prover/src/components/memory.rs
```diff
@@ -32,88 +34,142 @@ pub struct InteractionClaim {
...
+        Self {
+            log_size,
+            inputs: memory_boundaries,
+        }
+    }
+
     pub fn log_sizes(&self) -> TreeVec<Vec<u32>> {
-        let trace_log_sizes = vec![self.log_size; 3];
-        let interaction_log_sizes = vec![self.log_size; SECURE_EXTENSION_DEGREE];
+        let trace_log_sizes = vec![self.log_size; 2 * N_M31_IN_MEMORY_ENTRY];
```
Eikix, line 60: why are we doing `2 * N_M31_IN_MEMORY_ENTRY` here?

crates/prover/src/components/memory.rs
```diff
@@ -32,88 +34,142 @@ pub struct InteractionClaim {
...
+            .chain(std::iter::repeat([M31::from(0); N_M31_IN_MEMORY_ENTRY]))
+            .take(1 << self.log_size)
+            .array_chunks::<N_LANES>()
+            .map(|chunk| {
+                std::array::from_fn(|x| PackedM31::from_array(std::array::from_fn(|y| chunk[y][x])))
+            })
+            .collect();
+
+        // Generate lookup data and fill the trace
+        let mut loookup_data_initial_memory = Vec::new();
```
Eikix, line 127: typo

crates/prover/src/components/memory.rs
```diff
@@ -130,15 +186,39 @@ impl InteractionClaim {
...
         let mut interaction_trace = LogupTraceGenerator::new(log_size);
 
         let mut col = interaction_trace.new_col();
-        (col.par_iter_mut(), &lookup_data.memory)
+        (col.par_iter_mut(), &lookup_data.initial_memory)
             .into_par_iter()
             .for_each(|(writer, value)| {
-                let denom: PackedQM31 = memory.combine(&value[0..2]);
-                writer.write_frac(-PackedQM31::one() * value[2], denom);
+                let denom: PackedQM31 = memory.combine(&value[..6]);
```
Eikix, line 199: why 6 here?

crates/prover/src/components/memory.rs
```diff
@@ -32,88 +34,142 @@ pub struct InteractionClaim {
...
     {
-        let mut counts: HashMap<(u32, u32), u32> = HashMap::new();
-        for entry in lookup_data {
-            let arr0 = entry[0].to_array();
-            let arr1 = entry[1].to_array();
-            for (val0, val1) in arr0.iter().zip(arr1.iter()) {
-                let key = (val0.0, val1.0);
-                *counts.entry(key).or_insert(0) += 1;
+        // Pack memory entries from the prover input
+        let initial_memory: Vec<[PackedM31; N_M31_IN_MEMORY_ENTRY]> = self
```
Eikix, line 79: pardon my ignorance but what is inside the trace of the memory component? it looks like we only 'log' the memory at the initial step and the final step?

are we not supposed to log every single memory access?

crates/prover/src/components/memory.rs
ClementWalter: this is the memory component for initial and final values, ie boundary conditions. So it's just a list of entries with appropriate multiplicity (-1 or +1)

crates/prover/src/components/memory.rs
ClementWalter: opcode components "clear" and "write" memory values, the memory component is only useful for boundary conditions indeed

Eikix: and checking that "squashing" iterative prev_value -> value is consistent?

crates/prover/src/components/memory.rs
```diff
@@ -1,27 +1,28 @@
...
 use stwo_prover::core::pcs::TreeVec;
-use stwo_prover::core::poly::circle::{CanonicCoset, CircleEvaluation};
+use stwo_prover::core::poly::circle::CircleEvaluation;
 use stwo_prover::core::poly::BitReversedOrder;
 
+use crate::adapter::memory::Memory;
 use crate::relations;
 
-#[derive(Copy, Clone, Default)]
+const N_M31_IN_MEMORY_ENTRY: usize = 7;
```
ClementWalter, line 23:
```suggestion
// address, value, clock, multiplicity
const N_M31_IN_MEMORY_ENTRY: usize = 7;
```

crates/prover/src/components/memory.rs
```diff
@@ -137,8 +150,14 @@ impl InteractionClaim {
...
             .for_each(|(writer, value)| {
-                let denom: PackedQM31 = memory.combine(&value[0..2]);
-                writer.write_frac(-PackedQM31::one() * value[2], denom);
+                let denom: PackedQM31 = memory.combine(&value[..6]);
+                let mult: PackedQM31 = PackedQM31::from_packed_m31s([
+                    value[6],
+                    PackedM31::broadcast(M31::from(0)),
+                    PackedM31::broadcast(M31::from(0)),
+                    PackedM31::broadcast(M31::from(0)),
+                ]);
```
ClementWalter, line 159: `impl From<PackedM31> for PackedQM31 {` does exist already

crates/prover/src/components/memory.rs
```diff
@@ -31,13 +32,24 @@ pub struct InteractionClaim {
...
 }
 
+#[derive(Uninitialized, IterMut, ParIterMut)]
 pub struct LookupData {
-    pub memory: Vec<[PackedM31; 3]>,
+    pub memory: Vec<[PackedM31; N_M31_IN_MEMORY_ENTRY]>,
 }
 
 impl Claim {
+    pub fn new(memory: &Memory) -> Self {
```
ClementWalter, line 41: don't get the memory in the `new` and keep the update from `write_trace`

AntoineFONDEUR: we need it to compute the log_size

ClementWalter: with default it's set to 0, then updated when writing the trace with

```
self.log_size = addresses.len().ilog2() + LOG_N_LANES;
```

crates/prover/src/components/memory.rs
```diff
@@ -31,13 +35,18 @@ pub struct InteractionClaim {
...
+#[derive(Uninitialized, IterMut, ParIterMut)]
 pub struct LookupData {
-    pub memory: Vec<[PackedM31; 3]>,
+    pub memory: Vec<[PackedM31; N_M31_IN_MEMORY_ENTRY]>,
 }
 
 impl Claim {
+    pub const fn new() -> Self {
+        Self { log_size: 0 }
+    }
```
ClementWalter, line 46: it's `Default`

crates/prover/src/components/mod.rs
```diff
@@ -175,34 +139,10 @@ impl<const N: usize> Components<N> {
...
-                },
-                interaction_claim
-                    .single_constraint_with_relation
-                    .claimed_sum,
-            ),
             memory: memory::Component::new(
                 location_allocator,
                 memory::Eval {
-                    claim: claim.memory,
+                    claim: claim.memory.clone(),
```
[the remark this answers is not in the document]
AntoineFONDEUR: In prove_cairo, Proof contains the claim so memory::Eval can't consume it hence the clone which is inexpensive in this case

ClementWalter (approved):

PR #77, opened 2025-06-19
fix(compiler): add support for immediate left operands

gael-bigot:
Closes CORE-858


crates/compiler/codegen/src/builder.rs
```diff
@@ -296,11 +288,74 @@ impl CasmBuilder {
...
+                            .with_off0(right_off)
+                            .with_off2(dest_off)
+                            .with_imm(*imm)
+                            .with_comment(format!(
+                                "[fp + {dest_off}] = [fp + {right_off}] op {imm}"
+                            ));
+                        self.instructions.push(instr);
+                    }
+                    // For subtraction and division, we store the immediate in a temporary variable
+                    // In the future we should add opcodes imm_fp_sub and imm_fp_div
```
enitrat, line 307:
```suggestion
                    // For subtraction and division, we store the immediate in a temporary variable
                    // TODO In the future we should add opcodes imm_fp_sub and imm_fp_div
```

enitrat (approved):

PR #79, opened 2025-06-19
feat(vm): add program length to memory trace

obatirou:


enitrat (dismissed):

zmalatrax (approved):

crates/runner/src/vm/mod.rs
```diff
@@ -109,8 +109,8 @@ impl TryFrom<Program> for VM {
             .collect();
 
         // Create memory and load instructions starting at address 0
-        let program_length = M31(qm31_instructions.len() as u32);
-        let final_pc = program_length;
+        let program_length = qm31_instructions.len() as u32;
+        let final_pc = M31(program_length);
```
zmalatrax, line 113: `program_length` must be M31, otherwise `final_pc` will revert

I don't expect programs to be more > 2^31 instructions, but still a limitation in the current design...

PR #82, opened 2025-06-19
feat(compiler): MIR lowering for loop/while loops

enitrat:


crates/compiler/mir/src/ir_generation/tests/snapshots/cairo_m_compiler_mir__ir_generation__tests__mir_generation_tests__control_flow_infinite_loop.snap
```diff
@@ -0,0 +1,53 @@
...
+      store %0, 0
+      jump 1
+
+    1:
+      %1 = %0 Add 1
+      store %0, %1
+      if %0 Eq 10 then jump 3 else jump 4
+
+    2:
+      return %0
```
gael-bigot, line 43: The exit block should be at the very end of the loop structure for readability

enitrat: so I checked how we could make this, unfortunately this complexifies the IR generation step as we need to keep track of the sequentiality of blocks, thus I'm in favor of keeping it this way

PR #84, opened 2025-06-20
feat(compiler): support comparison operators (no CASM lowering)

enitrat:


gael-bigot:
We may as well add support for unary `not` (!). Otherwise lgtm.

PR #88, opened 2025-06-23
bench(vm): use compiled file

obatirou:
Close CORE-899

### DIFF
--- a/crates/runner/benches/vm_benchmark.rs
+++ b/crates/runner/benches/vm_benchmark.rs
@@ -1,79 +1,42 @@
+use std::fs;
 use std::time::Duration;
 
-use cairo_m_common::{Instruction, Opcode, Program};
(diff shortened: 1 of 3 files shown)


enitrat (dismissed):

crates/runner/benches/vm_benchmark.rs
```diff
@@ -1,79 +1,42 @@
...
-        // Jump back to condition check
-        Instruction::new(
-            Opcode::JmpAbsImm,
-            [M31::from(3), Zero::zero(), Zero::zero()],
-        ), // jmp_abs_imm: jmp abs 3
-    ];
-
-    instructions
-}
+const FP_OFFSET: u32 = 3;
```
zmalatrax, line 10: Add some doc to explain what it corresponds to (specifically 0 arg, 1 return value) as it's not explicit with the current const name

zmalatrax (approved):

PR #93, opened 2025-06-24
feat(prover): store_imm component

AntoineFONDEUR:
closes CORE-825


crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,456 @@
...
+    pub padding_offset: usize,
+}
+impl Enabler {
+    pub const fn new(padding_offset: usize) -> Self {
+        Self { padding_offset }
+    }
+
+    pub fn packed_at(&self, vec_row: usize) -> PackedM31 {
+        let row_offset = vec_row * N_LANES;
+        if self.padding_offset <= row_offset {
```
Eikix, line 48:
```suggestion
        if row_offset >= self.padding_offset {
```

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        let M31_1 = PackedM31::broadcast(M31::from(1));
+        let enabler_col = Enabler::new(1);
+
+        trace
+            .par_iter_mut()
+            .zip(packed_inputs.par_iter())
+            .zip(lookup_data.par_iter_mut())
+            .for_each(|((mut row, input), lookup_data)| {
+                *row[0] = enabler_col.packed_at(1);
```
Eikix, line 128: why packed_at(1)?

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,456 @@
...
+                // Current registers
+                let pc = input.pc;
```
Eikix, line 130: you don't need to initialize pc, just do row[1] = input.pc

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,456 @@
...
+
+                // Memory read: instruction
+                let mem0_address = input.mem0_address;
+                let mem0_prev_val_0 = input.mem0_prev_val_0;
+                let mem0_prev_val_1 = input.mem0_prev_val_1;
+                let mem0_prev_val_2 = input.mem0_prev_val_2;
+                let mem0_prev_val_3 = input.mem0_prev_val_3;
+                let mem0_prev_clock = input.mem0_prev_clock;
+                *row[3] = mem0_address;
+                *row[4] = mem0_prev_val_0;
```
Eikix, line 144: you don't need `prev_value` since it never changes

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,456 @@
...
+                let mem1_prev_clock = input.mem1_prev_clock;
+                *row[14] = mem1_address;
+                *row[15] = mem1_prev_val_0;
+                *row[16] = mem1_prev_val_1;
+                *row[17] = mem1_prev_val_2;
+                *row[18] = mem1_prev_val_3;
+                *row[19] = mem1_prev_clock;
+                *lookup_data.memory[2] = [
+                    input.mem1_address,
+                    input.mem1_prev_val_0,
```
Eikix, line 191: you don't need prev_value if you're just reading

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,459 @@
...
+        ));
+
+        // Verify that imm = mem0_value_1
+        let imm = eval.next_trace_mask();
+        eval.add_constraint(imm - mem0_value_1);
+
+        // Final state verification
+        let new_pc = pc + E::F::one();
+        let new_pc_value = eval.next_trace_mask();
+        eval.add_constraint(new_pc - new_pc_value.clone());
```
Eikix, line 441:
```suggestion
        let new_pc = eval.next_trace_mask();
        eval.add_constraint(pc + E::F::one() - new_pc.clone());
```

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,456 @@
...
+                    input.mem1_address,
+                    input.mem1_value_0,
+                    input.mem1_value_1,
+                    input.mem1_value_2,
+                    input.mem1_value_3,
+                    input.mem1_clock,
+                ];
+
+                // Memory write: [fp + off2] = imm
+                let imm = input.mem0_value_1;
```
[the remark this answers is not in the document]
AntoineFONDEUR: I think we can delete it since we already added a column for it

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,456 @@
...
+                *row[25] = imm;
+
+                // Next registers
+                let new_pc = pc + M31_1;
+                *row[26] = new_pc;
+                *row[27] = fp;
+                *lookup_data.registers[1] = [new_pc, fp];
```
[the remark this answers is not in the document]
AntoineFONDEUR: We could but we won't be able to do it for all opcodes (for instance when there is a non trivial (pc,fp) update that require intermediate values to compute the update). Do we want to do it if it won't be consistent over all opcodes ?

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,459 @@
...
+        self.claim.log_size
+    }
+
+    fn max_constraint_log_degree_bound(&self) -> u32 {
+        self.log_size() + 1
+    }
+
+    #[allow(non_snake_case)]
+    fn evaluate<E: EvalAtRow>(&self, mut eval: E) -> E {
+        let M31_1 = E::F::from(M31::from(1));
```
ClementWalter, line 344: One actually needs the following collumns

// To cancel out relation padded values
- enabler
// The current registers
- pc
- fp
// The current clock for the rw memory
- clk
// The instruction value
- inst_val_0 // opcode id
- inst_val_1 // off0
- inst_val_2 // off1
- inst_val_3 // off2
- inst_prev_clk
// The dst value
- dst_prev_val
- dst_prev_clk

Then one does

// enabler is a bool
1. enabler * (1 - enabler)

// registers update is regular
2. -[pc, fp] +[pc + 1, fp] in Registers relation

// read instruction from memory
3. -[pc, inst_val_0, inst_val_1, inst_val_2, inst_val_3, inst_prev_clk] + [pc, inst_val_0, inst_val_1, inst_val_2, inst_val_3, clk] in Memory relation

// write off0 in [fp + off2]
4. -[fp + off2, dst_prev_val, 0, 0, 0, dst_prev_clk] + [fp + off2, off0, 0, 0, 0, clk] in Memory relation

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,306 @@
...
+        Self,
+    ) {
+        let log_size = lookup_data.memory[0].len().ilog2() + LOG_N_LANES;
+        let mut interaction_trace = LogupTraceGenerator::new(log_size);
+        let enabler_col = Enabler::new(non_padded_length);
+
+        let mut col0 = interaction_trace.new_col();
+        for (i, (value0, value1)) in lookup_data.registers[0]
+            .iter()
+            .zip(&lookup_data.memory[0])
```
ClementWalter, line 160: missing `lookup_data.registers[1]`

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,314 @@
...
+            .zip(&claim_data.lookup_data.memory[3])
+            .enumerate()
+        {
+            let denom_0: PackedQM31 = memory_relation.combine(value0);
+            let mult_0: PackedQM31 = -PackedQM31::from(enabler_col.packed_at(i));
+            let denom_1: PackedQM31 = memory_relation.combine(value1);
+            let mult_1: PackedQM31 = PackedQM31::from(enabler_col.packed_at(i));
+
+            col3.write_frac(i, mult_0 * denom_0 + mult_1 * denom_1, denom_0 * denom_1);
+        }
```
obatirou, line 211: This loop could be abstracted into a function + why go from col1 to col3?
Shouldn't this be col2?

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,314 @@
...
+                *row[10] = dst_prev_clock;
+                *lookup_data.memory[2] = [
+                    input.fp + off0,
+                    dst_prev_clock,
+                    dst_prev_val,
+                    zero,
+                    zero,
+                    zero,
+                ];
+                *lookup_data.memory[3] = [input.fp + off0, clock, off2, zero, zero, zero];
```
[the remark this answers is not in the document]
AntoineFONDEUR: typo indeed

crates/prover/src/components/store_imm.rs
AntoineFONDEUR: typo again thx

crates/prover/src/components/mod.rs
```diff
@@ -24,47 +23,65 @@ use crate::relations;
...
 
     pub fn mix_into(&self, channel: &mut impl Channel) {
+        self.store_imm.mix_into(channel);
         self.memory.mix_into(channel);
         self.range_check_20.mix_into(channel);
     }
 
     pub fn write_trace<MC: MerkleChannel>(
-        input: ProverInput,
+        mut input: ProverInput,
```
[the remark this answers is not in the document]
Eikix: i don't think that's the reason of the inputs being mutable,

it is because inside the write_trace function, we call input.resize(next_power_of_two) to pad the inputs.

if we want, we can pad the inputs inside the adapter, and make the non_padded_length available somehow

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,354 @@
...
+        // Update registers
+        eval.add_to_relation(RelationEntry::new(
+            &self.registers,
+            -E::EF::from(enabler.clone()),
+            &[pc.clone(), fp.clone()],
+        ));
+        eval.add_to_relation(RelationEntry::new(
+            &self.registers,
+            E::EF::from(enabler.clone()),
+            &[pc.clone() + E::F::one(), fp.clone()],
```
obatirou, line 291: Why not use the variable one defined above ?
Not sure E::F::from(M31::from(1)) and E::F::one() are equivalent?

crates/prover/src/components/store_imm.rs
```diff
@@ -0,0 +1,354 @@
...
+
+        // Check the write at fp + off2
+        eval.add_to_relation(RelationEntry::new(
+            &self.memory,
+            -E::EF::from(enabler.clone()),
+            &[
+                fp.clone() + off2.clone(),
+                dst_prev_clock.clone(),
+                dst_prev_val,
+            ],
```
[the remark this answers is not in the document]
AntoineFONDEUR: no stwo will fill the entry with zeros which is what we want

PR #95, opened 2025-06-24
feat(prover): serialize proof struct

zmalatrax:
Closes CORE-914


ClementWalter (approved):
just wondering why we sometimes use serde and somethims sonic

PR #96, opened 2025-06-24
feat(prover): Range check minor refacto

ClementWalter:
- **Create a mod for range_check and make sure rc_20 hardcodes 20 everywhere**
- **Add bench for range_check init method**


Eikix (approved):
lgtm

PR #101, opened 2025-06-24
refactor(vm-prover): unify trace entry types

zmalatrax:
Closes Core-917

### DIFF
--- a/crates/prover/src/adapter/mod.rs
+++ b/crates/prover/src/adapter/mod.rs
@@ -121,9 +121,9 @@ pub fn import_from_runner_output(
     let _span = span!(Level::INFO, "import_from_runner_output").entered();
 
     let vm = &runner_output.vm;
-    let trace_iter = vm.trace.iter().cloned();
+    let trace_iter = vm.trace.iter().copied();
     let memory_trace = vm.memory.trace.borrow();
-    let memory_iter = memory_trace.iter().cloned();
+    let memory_iter = memory_trace.iter().copied();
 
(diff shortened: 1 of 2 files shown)


crates/runner/src/memory/mod.rs
```diff
@@ -42,7 +37,7 @@ pub struct Memory {
     /// allows methods with immutable `&self` receivers, like `get_data`, to
     /// modify the trace. This design choice separates the logical immutability
     /// of an operation from the implementation detail of tracing.
-    pub trace: RefCell<Vec<TraceEntry>>,
+    pub trace: RefCell<Vec<MemoryEntry>>,
```
obatirou, line 40: nit: 
Should an alias be used either in runner on in adapter?
seems a little off compared to pushing a traceEntry here

crates/prover/src/adapter/mod.rs
```diff
@@ -137,9 +121,9 @@ pub fn import_from_runner_output(
     let _span = span!(Level::INFO, "import_from_runner_output").entered();
 
     let vm = &runner_output.vm;
-    let trace_iter = vm.trace.iter().map(|e| (e.pc, e.fp));
+    let trace_iter = vm.trace.iter().cloned();
     let memory_trace = vm.memory.trace.borrow();
-    let memory_iter = memory_trace.iter().map(|e| (e.addr, e.value));
+    let memory_iter = memory_trace.iter().cloned();
```
obatirou, line 126: Should Copy be used instead?

obatirou (approved):

PR #108, opened 2025-06-25
feat(vm-prover): provide final state to the prover

obatirou:
Close CORE-922


AntoineFONDEUR (dismissed):

zmalatrax (changes requested):
simply add the final state to the trace once the while loop is finished on the main interpreter loop of the runner

zmalatrax (approved):

PR #111, opened 2025-06-25
feat(prover): add relation tracker

AntoineFONDEUR:
close CORE-852


crates/prover/src/debug_tools.rs
ClementWalter, line 1: remove the `cairo` from the names

PR #112, opened 2025-06-25
optimize(vm): opcode fn dispatcher

enitrat:


enitrat:
sounds actually useless :) will close

PR #117, opened 2025-06-25
fix(prover): Enabler inequalities

ClementWalter:
- **Fix inequality in enabler**
- **Update .gitignore**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 **/test_data/*.json
+.DS_Store


enitrat (approved):

crates/prover/src/utils.rs
```diff
@@ -19,10 +19,10 @@ impl Enabler {
 
     pub fn packed_at(&self, vec_row: usize) -> PackedM31 {
         let row_offset = vec_row * N_LANES;
-        if row_offset <= self.padding_offset {
+        if row_offset >= self.padding_offset {
             return PackedM31::zero();
         }
-        if row_offset + N_LANES <= self.padding_offset {
+        if row_offset + N_LANES < self.padding_offset {
```
zmalatrax, line 25: Why replacing the `<=` with `<` ?

If `row_offset + N_LANES == self.padding_offset `, then the "partially enabled) is actually taking `self.padding_offset - offset == N_LANES`, which is fully enabled..

PR #118, opened 2025-06-26
feat(compiler): support multiple return values as tuples

enitrat:


gael-bigot:
This fails at codegen:
```
func foo(x: felt) -> (felt, felt){
    return (x, 1);
}

func main(){
    let result = foo(2);
    return result;
}
```
We should either fully implement tuples or modify mir to avoid using load and get_element_ptr

PR #120, opened 2025-06-26
feat(prover): improve memory footprint during adapter step & trace generation

Eikix:
resolve CORE-933


Eikix:
> do we have benchmarks on mem usage?

investigating

enitrat (approved):

PR #128, opened 2025-06-27
soundness(prover): Split relation column using mul

ClementWalter:


crates/prover/src/components/opcodes/store_mul_fp_fp.rs
```diff
@@ -315,18 +315,27 @@ impl InteractionClaim {
...
+                let numerator = -PackedQM31::from(enabler_col.packed_at(i));
+                let denom: PackedQM31 = memory_relation.combine(memory_prev);
 
-                let numerator = num_prev * denom_new + num_new * denom_prev;
-                let denom = denom_prev * denom_new;
+                writer.write_frac(numerator, denom);
+            });
+        col.finalize_col();
+
+        let mut col = interaction_trace.new_col();
```
AntoineFONDEUR, line 329: creating a new column here so you have to update N_LOOKUPS_COLUMNS

PR #132, opened 2025-07-01
feat(prover): add multiplicity to writes

AntoineFONDEUR:
Closes CORE-943


crates/prover/tests/prover.rs
ClementWalter, line 1: not sure why we remove these two tests

crates/prover/tests/prover.rs
```diff
@@ -89,8 +35,10 @@ fn test_prove_and_verify_fibonacci_program() {
         Default::default(),
     )
     .unwrap();
-
+    dbg!(&compiled_fib.program.instructions);
```
ClementWalter, line 38: remove dbg!

crates/prover/src/debug_tools/relation_tracker.rs
ClementWalter, line 1: don't remove `        add_to_relation_entries(&opcodes.jnz_fp_fp_taken, trace),`

AntoineFONDEUR:
This is not sound as one could very well read from the write preceding the previous write.

PR #139, opened 2025-07-08
feat(prover): prepare inputs to Merkle component

AntoineFONDEUR:
Closes CORE-962


crates/prover/src/adapter/mod.rs
```diff
@@ -88,7 +107,8 @@ pub fn import_from_runner_artifacts(
     let memory_file_iter = MemoryEntryFileIter::try_from(mem_path)?;
     let memory_iter = memory_file_iter.map(Into::into);
 
-    import_internal(trace_iter, memory_iter)
+    // Todo: serialize the initial memory
```
ClementWalter, line 110:
```suggestion
    unimplemented!("serialize the initial memory")
```

crates/prover/src/adapter/partial_merkle.rs
ClementWalter, line 1: can we make the hash function a generic of the lib ? (much like the merkel channel doesn't depend on a given hash function)

crates/prover/src/adapter/partial_merkle.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    defaults
+}
+
+static DEFAULT_HASHES: OnceLock<Vec<QM31>> = OnceLock::new();
+
+fn get_default_hashes() -> &'static Vec<QM31> {
+    DEFAULT_HASHES.get_or_init(compute_default_hashes)
+}
```
ClementWalter, line 58: make this a const to make sure the default hashes are computed at compile time

AntoineFONDEUR: `poseidon_hash` isn't const so it complicates things. Also I think we should just store the hash values in a plain hard-coded way in a const file once we have the real hash function.

ClementWalter: what prevents from making it a const ?

crates/prover/src/adapter/memory.rs
```diff
@@ -196,21 +197,41 @@ where
...
-                let initial_value = (memory_entry.value, M31::zero());
-                self.initial_memory
-                    .insert(memory_entry.address, initial_value);
-                initial_value
-            });
+            .unwrap_or_else(|| (memory_entry.value, M31::zero(), -M31::one()));
+
+        // If it's the first time we use a memory cell,
+        // We insert it in the initial memory with multiplicity 1.
+        // Thus we extend the initial memory (initial from the VM point of view) with first accesses.
```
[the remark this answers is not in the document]
AntoineFONDEUR: We take the initial memory from the VM and extend it with all addresses used during the program

crates/prover/src/adapter/partial_merkle.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    fn default_hashes() -> &'static [M31] {
+        use std::sync::OnceLock;
+        static DEFAULT_HASHES: OnceLock<Vec<M31>> = OnceLock::new();
+
+        DEFAULT_HASHES.get_or_init(|| {
+            let mut defaults = vec![M31::zero(); (TREE_HEIGHT + 1) as usize];
+
+            // Layer 0: hash of zero
+            defaults[0] = Self::hash(M31::zero(), M31::zero());
```
ClementWalter, line 46: I think that the initial value is actually `0` and hash starts at index `1`

crates/prover/src/adapter/partial_merkle.rs
```diff
@@ -0,0 +1,262 @@
...
+            defaults
+        })
+    }
+}
+
+/// Build a partial Merkle tree from a memory state
+/// Each QM31 value is split into 4 M31 leaves
+/// The tree has 31 layers (0 to 30):
+/// - Layer 0: Leaf layer with up to 2^30 M31 values (from 2^28 QM31 memory cells)
+/// - Layer 30: Root layer with a single hash value
```
ClementWalter, line 63:
```suggestion
/// - Layer 0: Root layer with a single hash value
/// - Layer 30: Leaf layer with up to 2^30 M31 values (from 2^28 QM31 memory cells)
```

crates/prover/src/adapter/partial_merkle.rs
```diff
@@ -0,0 +1,262 @@
...
+        // Process all nodes at this layer
+        let mut indices_to_process: Vec<u32> = current_layer_nodes.keys().copied().collect();
+        indices_to_process.sort_unstable();
+
+        let mut processed_indices = std::collections::HashSet::new();
+
+        for &index in &indices_to_process {
+            if processed_indices.contains(&index) {
+                continue;
+            }
```
ClementWalter, line 117: how can this be possible since indexes are from a hashmap ?

crates/prover/tests/fibonacci_trace_test.rs
```diff
@@ -34,6 +34,7 @@ pub mod fibonacci {
         use tempfile::TempDir;
 
         #[test]
+        #[ignore]
```
ClementWalter, line 37: ?

PR #147, opened 2025-07-10
refactor(compiler): use statements for imports

enitrat:


gael-bigot:
LSP is disabled when switching files after an error.
# Description
When the LSP is running in Cursor, if a linter error is produced while working on a .cm file, switching to another file tab will disable LSP for the second tab. This happens even if the error was corrected before switching tabs. 
# Reproduction
- Open the demo project
- Go to one of the files (eg. `math.cm`) and create a linter error (eg remove a semicolon)
- Switch to another file (eg. `main.cm`), and attempt to produce another error, nothing happens
- If you go back to the first tab LSP should work fine there

enitrat:
@gael-bigot can you try again after re-building LS and extension now?

PR #149, opened 2025-07-10
feat(prover): add the merkle component

AntoineFONDEUR:
closes CORE-977


crates/prover/src/components/memory.rs
```diff
@@ -128,34 +174,84 @@ impl InteractionClaim {
...
+            col.par_iter_mut(),
+            &interaction_claim_data.lookup_data.memory[0],
+            &interaction_claim_data.lookup_data.merkle[0],
+        )
+            .into_par_iter()
+            .enumerate()
+            .for_each(|(i, (writer, value0, value1))| {
+                let num0: PackedQM31 = PackedQM31::from(value0[6]);
+                let denom0: PackedQM31 = memory.combine(&value0[..6]);
+                let num1: PackedQM31 = -PackedQM31::from(enabler_col.packed_at(i));
```
ClementWalter, line 198:
```suggestion
                let num1: PackedQM31 = PackedQM31::from(enabler_col.packed_at(i));
```

crates/prover/src/components/memory.rs
```diff
@@ -128,34 +174,84 @@ impl InteractionClaim {
...
+            &interaction_claim_data.lookup_data.merkle[1],
+            &interaction_claim_data.lookup_data.merkle[2],
+        )
+            .into_par_iter()
+            .enumerate()
+            .for_each(|(i, (writer, value0, value1))| {
+                let num0: PackedQM31 = -PackedQM31::from(enabler_col.packed_at(i));
+                let denom0: PackedQM31 = merkle.combine(value0);
+                let num1: PackedQM31 = -PackedQM31::from(enabler_col.packed_at(i));
+                let denom1: PackedQM31 = merkle.combine(value1);
```
ClementWalter, line 220:
```suggestion
                let num0: PackedQM31 = PackedQM31::from(enabler_col.packed_at(i));
                let denom0: PackedQM31 = merkle.combine(value0);
                let num1: PackedQM31 = PackedQM31::from(enabler_col.packed_at(i));
                let denom1: PackedQM31 = merkle.combine(value1);
```

crates/prover/src/components/memory.rs
```diff
@@ -128,34 +174,84 @@ impl InteractionClaim {
...
+            &interaction_claim_data.lookup_data.merkle[3],
+        )
             .into_par_iter()
-            .for_each(|(writer, value)| {
-                let denom: PackedQM31 = memory.combine(&value[..6]);
-                let mult: PackedQM31 = PackedQM31::from(value[6]);
-                writer.write_frac(mult, denom);
+            .enumerate()
+            .for_each(|(i, (writer, value))| {
+                let numerator: PackedQM31 = -PackedQM31::from(enabler_col.packed_at(i));
```
ClementWalter, line 237:
```suggestion
                let numerator: PackedQM31 = PackedQM31::from(enabler_col.packed_at(i));
```

crates/prover/src/adapter/memory.rs
```diff
@@ -67,9 +68,9 @@ struct MemoryArg {
 // TODO: Memory Value can take a value enum(M31, QM31) instead of QM31 to save space
 #[derive(Debug, Default, Eq, PartialEq, Clone)]
 pub struct Memory {
-    // (value, clock, multiplicity which is -1, 0 or 1)
-    pub initial_memory: HashMap<M31, (QM31, M31, M31)>,
-    pub final_memory: HashMap<M31, (QM31, M31, M31)>,
+    // (value, clock, multiplicity which is -1, 0 or 1, depth)
```
ClementWalter, line 71:
```suggestion
    // (addr, depth) => (value, clock, multiplicity which is -1, 0 or 1)
```

crates/prover/src/adapter/memory.rs
```diff
@@ -222,15 +228,24 @@ impl Memory {
...
+                    initial_memory_cell.0, memory_entry.value,
+                    "Initial memory value mismatch"
+                );
+                // Update the multiplicity to 1
+                initial_memory_cell.2 = M31::one();
+            } else {
+                let initial_memory_entry = (memory_entry.value, M31::zero(), M31::one());
+                self.initial_memory.insert(
+                    (memory_entry.address, M31::from(TREE_HEIGHT)),
+                    initial_memory_entry,
```
ClementWalter, line 246: let's just keep the `prev_memory_entry` as an `Option`. Then, if it's `None`, then it's the first time we encounter this address. We can assert than the `memory_entry` matches what we have in `initial_memory`, and reassign `prev_memory_entry` to be `initial_memory_entry`

crates/prover/src/adapter/merkle.rs
```diff
@@ -13,6 +13,17 @@ pub struct NodeData {
     pub value_right: M31,
 }
 
+impl NodeData {
```
ClementWalter, line 16: could we update `NodeData` to have `index: M31` and `depth: u8` ?

crates/prover/src/adapter/memory.rs
AntoineFONDEUR: This doesn't work: if we read a value that is already present in the initial memory we don't enter the unwrap_or_else (we only enter it for first stores).

crates/prover/src/components/memory.rs
```diff
@@ -156,6 +289,7 @@ impl InteractionClaim {
 pub struct Eval {
     pub claim: Claim,
     pub memory: relations::Memory,
+    pub merkle: relations::Merkle,
```
ClementWalter, line 292: get `relations` directly as in the opcodes

PR #151, opened 2025-07-11
refactor(prover): optimize array filling in Enabler::packed_at

reallesee:
Replace manual iteration with the more efficient `fill` method when populating the enabler array. This improves code readability and potentially performance by using Rust's optimized array filling operation instead of manual iteration with mutable references.


ClementWalter:
can you rebase please?

PR #161, opened 2025-07-16
feat: ignore underscore-prefixed variables in unused variable warnings

aso20455:
Close CORE-1092

Implement TODO item from scope_check.rs to ignore variables with underscore prefix in unused variable detection. This follows the common programming convention where variables starting with underscore are intentionally unused. Added test case to verify the behavior works correctly


enitrat (changes requested):
please restore import validity checks

enitrat:
You'll probably want to rebase on main and make sure your changes only impact unused variable warnings

PR #162, opened 2025-07-16
refactor(compiler): func -> fn

enitrat:
refactors the keyword for functions from `func` to `fn`. 

Rationale is that it's way easier for LLM codegen and I observed mis-generations using `fn` for `func`. At the era of AI, it makes more sense.


ClementWalter (approved):
Image

Image

🙈

PR #173, opened 2025-07-18
fix(compiler): better handling of unused variables

gael-bigot:
- variables which are assigned to but never read are correctly flagged as unused (CORE-920)
- unused struct literals are not allocated anymore
- right hand side of unused assignements are evaluated only if they contain a function call


enitrat:
is this still relevant?

PR #175, opened 2025-07-18
feat(prover): Consolidate JnzFpImm opcode implementation

ClementWalter:
## Summary
- Merged the two separate JnzFpImm components (jnz_fp_imm and jnz_fp_imm_taken) into a single unified component
- Simplified the implementation by adding a `taken` field that determines whether the jump is taken or not
- Reduced code duplication and improved maintainability by handling both cases in one component

## Changes
- Modified `jnz_fp_imm.rs` to handle both taken and not-taken cases with conditional logic
- Removed the separate `jnz_fp_imm_taken.rs` component entirely
- Updated the registers relation to conditionally update PC based on the `taken` field
- Adjusted constraints to ensure either `op0_val = 0` or `taken = 1`
- Updated all references and debug tools to use the single component

Close CORE-1035


crates/prover/src/components/opcodes/jnz_fp_imm.rs
```diff
@@ -330,12 +352,20 @@ impl FrameworkEval for Eval {
...
         let op0_val = eval.next_trace_mask();
+        let op0_val_inv = eval.next_trace_mask();
+        let taken = eval.next_trace_mask();
 
         // Enabler is 1 or 0
         eval.add_constraint(enabler.clone() * (one.clone() - enabler.clone()));
 
-        // Op0 val is 0
-        eval.add_constraint(enabler.clone() * op0_val.clone());
+        // Either op0_val = 0 or taken = 1
```
AntoineFONDEUR, line 361: To be precise this comment shouldn't be "either" since the following constraint only enforces an "or". The constraint would pass with op0_val=0 AND taken=1.

crates/prover/src/components/opcodes/jnz_fp_imm.rs
```diff
@@ -330,12 +352,20 @@ impl FrameworkEval for Eval {
...
+        eval.add_constraint(enabler.clone() * op0_val.clone() * (taken.clone() - one.clone()));
+
+        // taken = op0_val * op0_val_inv
+        eval.add_constraint(enabler.clone() * (taken.clone() - op0_val.clone() * op0_val_inv));
```
AntoineFONDEUR, line 365: This constraint takes care of the "either" part as with taken=1 and op0_val=0 this doesn't pass.

crates/prover/src/components/opcodes/jnz_fp_imm.rs
```diff
@@ -219,20 +232,29 @@ impl InteractionClaim {
...
+        let mut col = interaction_trace.new_col();
+        (
+            col.par_iter_mut(),
+            &interaction_claim_data.lookup_data.registers[1],
+        )
+            .into_par_iter()
+            .enumerate()
+            .for_each(|(i, (writer, registers))| {
+                let num = PackedQM31::from(enabler_col.packed_at(i));
+                let denom: PackedQM31 = relations.registers.combine(registers);
```
AntoineFONDEUR, line 255: Maybe I'm wrong but it seems like this saves us 3 columns: 
- adding a column in the main trace for new_pc;
- adding a constraint new_pc - pc.clone() + one.clone() + taken * (imm.clone() - one); 
- batch the register that no longer have multiplicity problem

PR #179, opened 2025-07-21
feat(compiler): Adapt Compiler Codegen for Variable-Sized Instructions

obatirou:
Close CORE-1026


.github/workflows/rust-tests.yml
```diff
@@ -5,8 +5,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
```
enitrat, line 9: required?

crates/compiler/codegen/tests/test_cases/variable_instructions/variable_sized_jump.cm
enitrat, line 1: no snaphost associated to this test?

enitrat (commented):
lgtm, just legacy methods are no longer relevant, but the snapshots all look good

crates/compiler/codegen/src/lib.rs
```diff
@@ -86,59 +81,61 @@ impl InstructionBuilder {
...
     pub fn with_imm(mut self, imm: i32) -> Self {
-        self.operand = Some(Operand::Literal(imm));
+        self.operands.push(Operand::Literal(imm));
         self
     }
 
     /// Set a label operand (convenience method)
     pub fn with_label(mut self, label: String) -> Self {
-        self.operand = Some(Operand::Label(label));
+        self.operands.push(Operand::Label(label));
```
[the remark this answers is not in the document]
obatirou: done

obatirou:
> lgtm, just legacy methods are no longer relevant, but the snapshots all look good

@enitrat shouldn't the CasmBuilder through generate_arithmetic_op handle u32 opcodes?

enitrat:
It depends if you consider it required in this PR or not. If just refactoring for dynamic operandi size then not necessarily

obatirou:
> It depends if you consider it required in this PR or not. If just refactoring for dynamic operandi size then not necessarily

ok let's do that in an other PR then 
Removing the variable_sized_jump test as it does not make sense without it

enitrat (approved):

PR #189, opened 2025-07-24
feat(prover): updated public memory

AntoineFONDEUR:
closes CORE-1076


crates/common/src/program.rs
```diff
@@ -15,6 +17,41 @@ pub struct EntrypointInfo {
...
+        }
+    }
+
+    /// Converts ranges to flat address vector for compatibility
+    pub fn to_flat_addresses(&self) -> Vec<M31> {
+        let mut addresses = Vec::new();
+        addresses.extend(self.program.clone().map(M31::from));
+        addresses.extend(self.input.clone().map(M31::from));
+        addresses.extend(self.output.clone().map(M31::from));
+        addresses
```
ClementWalter, line 51: can we do only one `map(M31::from)` ?

AntoineFONDEUR:
```suggestion
        let mut addresses = Vec::new();
        addresses.extend(self.program.clone());
        addresses.extend(self.input.clone());
        addresses.extend(self.output.clone());
        addresses.into_iter().map(M31::from).collect()
```

ClementWalter (changes requested):
as per our discussion, the public memory should emit both leaves through the merkle relation and memory values through the memory relation to make sure the given data is actually part of the memory root

PR #195, opened 2025-07-25
epic: support variable instruction size

obatirou:
Close CORE-986


crates/common/src/instruction.rs
```diff
@@ -1,133 +1,313 @@
...
+    };
 }
 
-impl From<&Instruction> for [M31; 4] {
-    fn from(instruction: &Instruction) -> Self {
-        (*instruction).into()
+// Define all instructions with their opcodes, memory accesses, fields, and sizes
+define_instruction!(
+    // Arithmetic operations
+    StoreAddFpFp = 0, 3, fields: [src0_off, src1_off, dst_off], size: 4;     // [fp + dst_off] = [fp + src0_off] + [fp + src1_off]
```
obatirou, line 208: It should use an Opcode enum instead of constants

crates/runner/src/vm/instructions/call.rs
```diff
@@ -62,7 +67,12 @@ pub fn call_abs_imm(
 ///
 /// CASM equivalent:
 /// `ret`
-pub fn ret(memory: &mut Memory, state: State, _: &Instruction) -> Result<State, MemoryError> {
+pub fn ret(
+    memory: &mut Memory,
+    state: State,
+    instruction: &Instruction,
+) -> Result<State, InstructionExecutionError> {
+    extract_as!(instruction, Ret);
```
obatirou, line 75:
```suggestion
```

crates/prover/src/adapter/memory.rs
```diff
@@ -33,28 +34,48 @@ impl From<crate::adapter::io::IoMemoryEntry> for MemoryEntry {
...
     pub operands: [Option<DataAccess>; 3],
 }
 
+impl Default for ExecutionBundle {
+    fn default() -> Self {
+        Self {
+            registers: VmRegisters::default(),
+            clock: M31::zero(),
+            instruction: InstructionAccess {
+                instruction: Instruction::Ret {},
```
enitrat, line 71: this is problematic as in store_fp_imm we do

```
        inputs.resize(1 << log_size, ExecutionBundle::default());
```

where we get operands[1] which is not defined.

enitrat (approved):

PR #199, opened 2025-07-28
feat(prover): add clock update component

AntoineFONDEUR:
closes CORE-1049

### DIFF
--- a/crates/prover/src/components/clock_update.rs
+++ b/crates/prover/src/components/clock_update.rs
@@ -1,0 +1,263 @@
+use num_traits::{One, Zero};
+use rayon::iter::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator};
+use serde::{Deserialize, Serialize};
+use stwo_air_utils::trace::component_trace::ComponentTrace;
+use stwo_air_utils_derive::{IterMut, ParIterMut, Uninitialized};
+use stwo_constraint_framework::logup::LogupTraceGenerator;
+use stwo_constraint_framework::{
+    EvalAtRow, FrameworkComponent, FrameworkEval, Relation, RelationEntry,
+};
+use stwo_prover::core::backend::BackendForChannel;
+use stwo_prover::core::backend::simd::SimdBackend;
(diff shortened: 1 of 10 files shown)


crates/prover/src/components/clock_update.rs
ClementWalter, line 1: add markdown description of the component

crates/prover/src/components/mod.rs
```diff
@@ -94,8 +101,17 @@ impl Claim {
...
+        let (clock_update_claim, clock_update_trace, clock_update_interaction_claim_data) =
+            clock_update::Claim::write_trace(&input.memory.clock_update_data);
+
         // Write range_check components
-        let range_check_data = opcodes_interaction_claim_data.range_check_20();
+        let opcodes_range_check_data = opcodes_interaction_claim_data.range_check_20();
+        let clock_update_range_check_data = clock_update_interaction_claim_data
+            .lookup_data
+            .range_check_20[0]
+            .par_iter();
```
ClementWalter, line 113: maybe add a function `range_check_20` to the `InteractionClaimData` to move this there and keep the code simpler here

crates/prover/src/components/clock_update.rs
ClementWalter, line 1: we can avoid the range check for the new clock and use directly the hard coded delta (=2**30 in the witness)

crates/prover/src/components/clock_update.rs
```diff
@@ -0,0 +1,268 @@
...
+//! - prev_clk
+//! - QM31 value
+//!
+//! # Constraints
+//!
+//! * enabler is a bool
+//!   * `enabler * (1 - enabler)`
+//! * update the clock
+//!   * `- [addr, prev_clk, value]` in `Memory` relation
+//!   * `+ [addr, inter_clk, value]` in `Memory` relation
```
ClementWalter, line 16:
```suggestion
//!   * `+ [addr, prev_clk + RC_20, value]` in `Memory` relation
```

PR #218, opened 2025-08-03
feat(compiler): ignore underscore-prefixed variables in unused variable warnings

aso20455:
Ignore variables with underscore prefix (_variable) in unused variable warnings. This is a common convention for intentionally unused variables
Added tests for both local variables and function parameters with underscore prefix


enitrat (approved):
LGTM!

PR #220, opened 2025-08-04
feat(runner-compiler): add u32_store_imm instruction

enitrat:


crates/common/tests/instruction_tests.rs
```diff
@@ -553,6 +553,20 @@ fn test_try_from_smallvec() {
             "U32StoreDivFpImm instruction",
         ),
         // Memory and control flow operations
+        (
```
zmalatrax, line 556: Comment should be moved down to `CallAbsImm`

enitrat: thought it would still count as memory op...

PR #225, opened 2025-08-06
doc(prover): add documentation for prover crate

AntoineFONDEUR:
closes CORE-927

### DIFF
--- a/crates/prover/src/relations.rs
+++ b/crates/prover/src/relations.rs
@@ -1,53 +1,31 @@
 #![allow(non_camel_case_types)]
-#![allow(clippy::unused_doc_comments)]
 use stwo_constraint_framework::relation;
 
-/// 20-bit range check relation for field arithmetic bounds.
-///
-/// Ensures values are within the valid range [0, 2^20).
-///
-/// ## Structure
-/// - **value**: The field element to range check
+// 20-bit range check relation for field arithmetic bounds.
+// Ensures values are within the valid range [0, 2^20).
+// Structure: value (the field element to range check)
 relation!(RangeCheck20, 1);
 
-/// Memory access relation for read/write operations.
-///
-/// Tracks all memory operations with address, clock, and QM31 values.
-///
-/// ## Structure
-/// - **addr**: Memory address (M31)
-/// - **clock**: Access timestamp (M31)
-/// - **value0-3**: QM31 value components (4 M31 elements)
+// Memory access relation for read/write operations.
+// Tracks all memory operations with address, clock, and QM31 values.
+// Structure: addr, clock, value0-3 (QM31 value components)
 relation!(Memory, 6);
 
-/// VM register state relation for PC and FP tracking.
-///
-/// Maintains consistency of program counter and frame pointer updates.
-///
-/// ## Structure
-/// - **pc**: Program counter (M31)
-/// - **fp**: Frame pointer (M31)
+// VM register state relation for PC and FP tracking.
+// Maintains consistency of program counter and frame pointer updates.
+// Structure: pc (program counter), fp (frame pointer)
 relation!(Registers, 2);
 
-/// Merkle tree node relation for memory commitments.
-///
-/// Ensures correct Merkle tree construction and hash computations.
-///
-/// ## Structure
-/// - **index**: Node index at tree level (M31)
-/// - **layer**: Tree depth/layer (M31)
-/// - **value**: Hash value at this node (M31)
-/// - **root**: Tree root hash (M31)
+// Merkle tree node relation for memory commitments.
+// Ensures correct Merkle tree construction and hash computations.
+// Structure: index (node index), layer (tree depth), value (hash), root (tree root hash)
 relation!(Merkle, 4);
 
-/// Poseidon2 hash function relation for cryptographic computations.
-///
-/// Connects the Poseidon2 component to the rest of the components: Poseidon2 component uses the initial
-/// state and emits the final digest. Other component that need to prove Poseidon2 computation should
-/// emit the initial state and comsume the final digest.
-///
-/// ## Structure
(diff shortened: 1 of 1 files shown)


crates/prover/src/adapter/merkle.rs
```diff
@@ -1,3 +1,19 @@
+//! # Merkle Tree Construction for Memory Commitments
+//!
+//! This module implements partial Merkle tree construction for Cairo-M memory states.
+//! The initial memory commited is vm.memory before the first step of a segment of execution.
+//! The final memory commited is vm.memory after the last step of a segment of execution.
+//! The memory commitments can contain memory values that are not accessed during the execution (ie not part of the
+//! memory log).
```
enitrat, line 7: what are commitments used in?

crates/prover/src/lib.rs
```diff
@@ -23,11 +39,27 @@ use stwo_prover::core::vcs::ops::MerkleHasher;
...
+///
+/// This structure contains all the necessary components to verify that a Cairo-M
+/// program was executed correctly, including:
+/// - Execution trace claims and proofs
+/// - Public input/output data
+/// - The underlying STARK proof
+/// - Proof-of-work nonce for additional security
+///
+/// ## Type Parameters
+/// * `H` - The Merkle hasher used for tree commitments (typically Blake2s)
```
enitrat, line 52: isnt it Poseidon?

crates/prover/src/adapter/memory.rs
```diff
@@ -37,31 +49,68 @@ impl From<crate::adapter::io::IoMemoryEntry> for MemoryEntry {
...
     pub prev_clock: M31,
 }
 
+/// Represents a complete execution step with all associated memory accesses.
+///
+/// An execution bundle contains:
+/// - The current register state (PC, FP)
+/// - The current clock time: clock is incremented at each step
+/// - The instruction being executed
+/// - Up to 3 operand memory accesses
```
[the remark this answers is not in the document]
AntoineFONDEUR: Yes atm it's only 3 operands per instruction, operands can however be u32 or felts.

crates/prover/src/relations.rs
```diff
@@ -1,15 +1,53 @@
...
-/// (~124 bits) + `INTERACTION_POW_BITS` - log2(number of relation terms).
-///
-/// E.g. assuming a 100-bit security target, the witness may contain up to
-/// 1 << (24 + INTERACTION_POW_BITS) relation terms.
+// 20-bit range check relation for field arithmetic bounds.
+//
+// Ensures values are within the valid range [0, 2^20).
+//
+// ## Structure
+// - **value**: The field element to range check
```
enitrat, line 9:
```suggestion
/// 20-bit range check relation for field arithmetic bounds.
///
/// Ensures values are within the valid range [0, 2^20).
///
/// ## Structure
/// - **value**: The field element to range check
```

comments or doc? (applicable below as well)

crates/prover/src/utils/enabler.rs
```diff
@@ -2,30 +2,75 @@ use num_traits::{One, Zero};
 use stwo_prover::core::backend::simd::m31::{PackedM31, N_LANES};
 use stwo_prover::core::fields::m31::M31;
 
-/// The enabler column is a column of length `padding_offset.next_power_of_two()` where
-/// 1. The first `padding_offset` elements are set to 1;
-/// 2. The rest are set to 0.
+/// Manages trace row enablement for constraint activation and padding.
+///
+/// The enabler creates a boolean column where the first `padding_offset` rows
```
enitrat, line 7: not sure I understand its purpose here

crates/prover/src/adapter/merkle.rs
AntoineFONDEUR: used in continuation: the prover commits to the initial and final memories

enitrat (commented):
great !

crates/prover/src/adapter/memory.rs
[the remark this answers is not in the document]
enitrat: not sure I follow you:

U32AddFpImm will need `src_offset, imm_lo, imm_hi, dst_offset` so 4 operands

crates/prover/src/poseidon2.rs
```diff
@@ -7,23 +7,9 @@ use zkhash::poseidon2::poseidon2_instance_m31::POSEIDON2_M31_16_PARAMS;
 
 use crate::adapter::merkle::MerkleHasher;
 
-/// ALPHA: power in the S-Box. The paper suggest x**5 as S-Box.
-///
-/// Note that stwo-cairo uses alpha=3.
-/// We pick 5 so that gcd(PRIME-1, 5) = 1 holds, indeed gcd(PRIME-1, 3) = 3.
-/// This const is exclusively used in the build script not for the hash computation nor the AIR (where x**5 s-box is hardcoded)
```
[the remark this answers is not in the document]
AntoineFONDEUR: yes it was unused and confusing since parameters are imported from the reference lib

crates/prover/src/adapter/merkle.rs
AntoineFONDEUR: sure 👍

crates/prover/src/relations.rs
AntoineFONDEUR: trunk was complaining about it, will fix it

crates/prover/src/adapter/memory.rs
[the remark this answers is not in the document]
AntoineFONDEUR: The advantage of having the range-based DataAccess is that it saves an address field per operand. And an U32 will always be split in two consecutive felts so the address is implicit, I don’t think it’s much of a problem.

PR #228, opened 2025-08-07
feat(compiler-runner): implement U32 comparison opcodes and fix return value slot calculation

enitrat:
## Summary
- Implemented U32 comparison opcodes (eq, neq, gt, ge, lt, le) that return felt values
- Fixed return value slot calculation to account for multi-slot types (U32 takes 2 slots)
- Fixed immediate operand ordering in U32 arithmetic instructions (imm_lo, imm_hi)

### New U32 Comparison Instructions
- Added new comparison opcodes: `U32StoreEqFpFp`, `U32StoreNeqFpFp`, `U32StoreGtFpFp`, `U32StoreGeFpFp`, `U32StoreLtFpFp`, `U32StoreLeFpFp` and imm variants
- These operations compare two U32 values and store the result as a felt (0 or 1)

### Codegen Improvements
- Updated binary operation handling to distinguish between arithmetic operations (return U32) and comparison operations (return felt)
- Fixed return value slot calculation to properly handle multi-slot types
- Added support for comparison operations with immediate values
- Corrected immediate operand ordering from (imm_hi, imm_lo) to (imm_lo, imm_hi) for consistency

### Test Updates
- Updated snapshot tests to reflect the new instruction formats and slot calculations


crates/runner/src/vm/instructions/store.rs
```diff
@@ -255,6 +256,68 @@ pub fn u32_store_imm(
...
+where
+    F: Fn(u32, u32) -> bool,
+{
+    let lhs = memory.get_u32(state.fp + src0_off)?;
+    let rhs = memory.get_u32(state.fp + src1_off)?;
+
+    let res = if op(lhs, rhs) {
+        QM31::from(M31::one())
+    } else {
+        QM31::from(M31::from(0))
```
zmalatrax, line 282:
```suggestion
        QM31::from(M31::zero())
```

zmalatrax (approved):

PR #248, opened 2025-08-25
Adapt AbiCodec for fixed-size arrays

enitrat:


crates/common/src/abi_codec.rs
```diff
@@ -521,21 +527,33 @@ mod tests {
...
+            let params = [AbiSlot {
+                name: "arr".into(),
+                ty: AbiType::FixedSizeArray {
+                    element: Box::new(AbiType::Felt),
+                    size: 3,
+                },
+            }];
+            let values = [InputValue::List(vec![0.into(), 1.into()])]; // only 2 provided
+            let err = encode_input_args(&params, &values).unwrap_err().to_string();
+            assert!(err.contains("array size mismatch"));
```
obatirou, line 556: Should the assert ensure the error is a `AbiCodecError::TypeMismatch` and not just assert the error string?

crates/runner/src/lib.rs
```diff
@@ -60,8 +60,217 @@ pub struct RunnerOutput {
...
+                    v
+                ))
+                .into());
+            }
+            Ok((CairoMValue::Bool(v == 1), 1))
+        }
+        AbiType::U32 => {
+            let lo = vm.memory.get_data(base)?;
+            let hi = vm.memory.get_data(base + M31::from(1u32))?;
+            if lo.0 >= (1 << 16) || hi.0 >= (1 << 16) {
```
obatirou, line 109: should constants be defined instead of magic numbers?
There might have other instances of this in the codebase. Maybe
```
const U16_MAX: u32 = 0xFFFF
if lo.0 > U16_MAX || hi.0 > U16_MAX
```

crates/runner/src/lib.rs
```diff
@@ -74,88 +283,208 @@ pub fn run_cairo_program(
...
+            if *size as usize != values.len() {
+                return Err(AbiCodecError::TypeMismatch(format!(
+                    "array size mismatch: expected {} got {}",
+                    size,
+                    values.len()
+                ))
+                .into());
+            }
+
+            // First, encode each element into a temporary buffer using the *argument ABI*.
```
obatirou, line 455: this method will be used in the hot pass (run_cairo_program) and temporary buffers / unecessary allocations should be avoided.

crates/runner/src/lib.rs
obatirou: it is obvious but for a reader there is a mental burden each time you encounter magic numbers you need to interpret. 
Anyway this is a nit.

crates/runner/tests/abi_codec_tests.rs
```diff
@@ -0,0 +1,525 @@
...
+    )
+    .expect("Failed to run program");
+
+    match &output.return_values[0] {
+        CairoMValue::Felt(v) => assert_eq!(v.0, 3, "Expected count of 3"),
+        _ => panic!("Expected Felt return value"),
+    }
+}
+
+#[ignore = "structs with arrays are not working well."]
```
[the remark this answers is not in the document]
enitrat: fixed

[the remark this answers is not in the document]
obatirou: Not pushed?

crates/runner/src/lib.rs
obatirou: Sure but there are 2 pass and create a temporary vec `elements_m31` that could be avoided by doing the encoding and writing to memory in only one pass.

enitrat: I don't see how you can avoid two loops here

PR #251, opened 2025-08-26
feat(prover): add u32 arithmetic components

AntoineFONDEUR:
closes CORE-1029
closes CORE-1133


crates/prover/src/components/opcodes/mod.rs
```diff
@@ -168,6 +172,38 @@ macro_rules! define_opcodes {
...
+                    .flatten()
+            )
+            .chain(
+                $self.u32_store_mul_fp_imm
+                    .lookup_data
+                    .range_check_16
+                    .par_iter()
+                    .flatten()
+            )
+    };
```
ClementWalter, line 205: Use `std::mem::discriminant` to check if a struct field exists in a macro by comparing field access. Here's a concise example:

```rust
macro_rules! check_value_field {
    ($struct_instance:expr) => {
        {
            let has_value = {
                #[allow(unused)]
                let temp = &$struct_instance;
                match std::mem::discriminant(&temp.value) {
                    _ => true,
                }
            };
            if has_value {
                println!("Value field exists: {}", $struct_instance.value);
            } else {
                println!("Value field does not exist");
            }
        }
    };
}

struct MyStruct1 {
    key: u32,
}

struct MyStruct2 {
    key: u32,
    value: u32,
}

fn main() {
    let s1 = MyStruct1 { key: 1 };
    let s2 = MyStruct2 { key: 1, value: 2 };

    check_value_field!(s1); // Prints: Value field does not exist
    check_value_field!(s2); // Prints: Value field exists: 2
}
```

This macro attempts to access the `value` field and uses a match to safely check its existence, avoiding compilation errors for structs without the field.

(courtesy of grok)

crates/prover/src/components/opcodes/u32_store_add_fp_imm.rs
```diff
@@ -154,53 +175,128 @@ impl Claim {
...
+                // Read op0_hi
+                *lookup_data.memory[6] = [
+                    fp + src_off + one,
+                    op0_prev_clock,
+                    op0_val_hi,
+                    zero,
+                    zero,
+                    zero,
+                ];
+                *lookup_data.memory[7] = [fp + src_off + one, clock, op0_val_hi, zero, zero, zero];
```
ClementWalter, line 265: iiuc, this is a mistake as `hi` is just the second limb at address `src_off` and not the first limb at the next address, since each address stores a QM31

crates/prover/src/components/opcodes/u32_store_add_fp_imm.rs
```diff
@@ -401,27 +593,41 @@ impl FrameworkEval for Eval {
 
     fn evaluate<E: EvalAtRow>(&self, mut eval: E) -> E {
         let one = E::F::from(M31::one());
-        let opcode_constant = E::F::from(M31::from(STORE_SUB_FP_FP));
+        let m31_16_shift = E::F::from(M31::from(1 << 16));
```
ClementWalter, line 596: `two_pow_16`

crates/prover/src/preprocessed/mod.rs
```diff
@@ -62,6 +62,6 @@ impl PreProcessedTraceBuilder {
 
 impl Default for PreProcessedTraceBuilder {
     fn default() -> Self {
-        Self::new().with_range_check(20)
+        Self::new().with_range_check(20).with_range_check(16)
```
ClementWalter, line 65: put in order

PR #260, opened 2025-08-27
Add compile-time check for opcode AIR implementation coverage

ClementWalter:
## Summary
This PR adds a compile-time check to ensure all opcodes defined in the `Instruction` enum have corresponding AIR implementations in the prover.

## Changes
- Modified the `define_opcodes!` macro to use opcode variant names directly instead of constants
- Added a compile-time exhaustiveness check that verifies all `Instruction` variants are handled
- Listed TODOs for opcodes not yet implemented in the AIR
- Simplified the macro by removing redundant const declarations

## Benefits
- Prevents accidentally forgetting to implement AIR for new opcodes
- Makes the macro more maintainable by reducing duplication
- Provides clear visibility of which opcodes still need AIR implementation

The compile-time check ensures that if a new opcode is added to the `Instruction` enum but not to the AIR implementation, the code will fail to compile with a clear error message.


enitrat (approved):
very good

PR #261, opened 2025-08-28
feat(prover): add range_check 8 and 16

AntoineFONDEUR:
closes CORE-1133


crates/prover/src/preprocessed/range_check/mod.rs
```diff
@@ -7,7 +7,13 @@ use stwo_prover::core::poly::BitReversedOrder;
 
 use crate::preprocessed::PreProcessedColumn;
 
+// Include the macro implementation
+#[macro_use]
+pub mod range_check_macro;
+
+pub mod range_check_16;
 pub mod range_check_20;
+pub mod range_check_8;
```
ClementWalter, line 16: you can just define them here, no need for a dedicated file

crates/prover/src/components/opcodes/mod.rs
```diff
@@ -216,5 +235,6 @@ define_opcodes!(
         ],
         store_fp_imm
     ),
-    ([const STORE_DOUBLE_DEREF_FP], store_double_deref_fp)
+    ([const STORE_DOUBLE_DEREF_FP], store_double_deref_fp),
```
ClementWalter, line 238: need to rebase

ClementWalter (commented):
if we cannot find a way to avoid filling manually the range_check 8 and 16 data, I guess that it's the sign that one needs to define actual traits for the components with a default empty list for these

ClementWalter (approved):

ClementWalter (approved):

PR #269, opened 2025-08-28
feat(prover): add u32_store_imm

AntoineFONDEUR:
closes CORE-1169


ClementWalter (approved):
tbf this component is just a memcpy of len 2. This will probably be dropped with memory refacto. Also, without a range_checked memory by design we'll end up range checking far too much

PR #277, opened 2025-08-29
feat (wasm): add fib compilation from rust source

gael-bigot:
Closes CORE-1149
Closes CORE-1148


crates/wasm/src/flattening.rs
```diff
@@ -667,30 +664,119 @@ impl DagToMir {
...
+                        result_id,
+                        Value::operand(temp2),
+                        Value::operand(temp1),
+                    ),
+                    Op::I32GtS => Instruction::binary_op(
+                        BinaryOp::U32Greater,
+                        result_id,
+                        Value::operand(temp1),
+                        Value::operand(temp2),
+                    ),
```
enitrat, line 738: MIR has 
```
    U32Less,
    U32Greater,
    U32LessEqual,
    U32GreaterEqual,
 ```
 adaption to VM-specific opcodes is done during lowering

Comment (gael-bigot) [reply crates/wasm/src/flattening.rs:738]: I realised afterwards that the change was codegen-specific. However we might benefit from converting to VM opcodes during the mir phase to benefit  as much as possible from optimisation passes.

File: crates/wasm/src/flattening.rs [hunk 2]
```diff
@@ -667,30 +664,119 @@ impl DagToMir {
...
+                wasm_op,
+                inputs[0],
+                inputs[1],
+                MirType::U32,
+                context,
+            ),
+
+            // For comparisons, we produce a boolean result
+            // This is not WASM compliant, but works if these values are only used in conditional branches
+            // TODO : cast everything correctly or sync with VM so that comparisons between u32 produce u32 booleans
```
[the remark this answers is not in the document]
gael-bigot: If the original WASM code keeps the result of an `I32Eq` for example and tries to add it to another value, this will result in a felt on U32 operation and crash the VM.

crates/wasm/src/flattening.rs
enitrat: you're lowering WASM to MIR, so you don't want to reason about anything that happens under that level

crates/wasm/src/flattening.rs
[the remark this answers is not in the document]
enitrat: either way it's true that it's bad because you can't emit to MIR without thinking about the underlying architecture which goes against my previous comment :/

enitrat (approved):

PR #287, opened 2025-09-01
fix(cargo-cairo-m): replace template placeholders with valid one

obatirou:
Close CORE-1187

Replace {{name}} with cairo-m-template in Cargo.toml template to avoid cargo compilation errors when parsing template files


zmalatrax (approved):
Looks fine to me

PR #297, opened 2025-09-04
refactor: remove AssertEqFpFp

enitrat:
Compile assertions on two operands by using `Eq` then AssertEqFpImm with `1`.


enitrat:
@codex Review this PR.

PR #298, opened 2025-09-04
feat(prover): add bitwise preprocessed column

AntoineFONDEUR:
closes CORE-1206


crates/prover/src/preprocessed/mod.rs
```diff
@@ -55,6 +56,16 @@ impl PreProcessedTraceBuilder {
...
 
+    pub fn with_bitwise_stacked(mut self) -> Self {
+        // Create 4 columns for the stacked bitwise operations
+        // (operation_id, input1, input2, result)
+        for col_index in 0..4 {
+            let bitwise = bitwise::BitwiseStacked::new(col_index);
+            self.columns.push(Box::new(bitwise));
+        }
+        self
+    }
```
ClementWalter, line 67:
```suggestion
    pub fn with_bitwise(mut self) -> Self {
        // Take as input the bit size of the bitwise operands
        self.columns.push(Bitwise.new(8).columns.map(|c| => Box::new(c));        
        self
    }
```

something like that

crates/prover/src/preprocessed/bitwise.rs
```diff
@@ -0,0 +1,357 @@
...
+            }
+        }
+
+        // Pad with zeros to reach power of 2
+        while op_id_col.len() < total_size {
+            op_id_col.push(M31(0));
+            input1_col.push(M31(0));
+            input2_col.push(M31(0));
+            result_col.push(M31(0));
+        }
```
ClementWalter, line 138: use `.resize(n, M31(0));`

crates/prover/src/preprocessed/bitwise.rs
```diff
@@ -0,0 +1,357 @@
...
+        // Pack data for interaction
+        let packed_data: Vec<[PackedM31; 5]> = (0..total_size)
+            .collect::<Vec<_>>()
+            .par_chunks(N_LANES)
+            .enumerate()
+            .map(|(chunk_idx, _chunk)| {
+                let base_idx = chunk_idx * N_LANES;
+                [
+                    PackedM31::from_array(std::array::from_fn(|i| {
+                        op_id_col.get(base_idx + i).copied().unwrap_or(M31(0))
```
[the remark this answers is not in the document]
AntoineFONDEUR: since we padded we could just use [] instead of get yes

crates/prover/src/preprocessed/bitwise.rs
```diff
@@ -0,0 +1,357 @@
...
+            BitwiseStacked::new(1, self.operand_bits),
+            BitwiseStacked::new(2, self.operand_bits),
+            BitwiseStacked::new(3, self.operand_bits),
+        ]
+    }
+}
+
+/// Stacked bitwise preprocessed column
+/// Stacks AND, OR, XOR operations into a single column
+pub struct BitwiseStacked {
```
ClementWalter, line 286: rename as `BitwiseCol` and remove `pub`. We don't want to see this outside of this file

crates/prover/src/preprocessed/bitwise.rs
```diff
@@ -0,0 +1,357 @@
...
+                    _ => unreachable!(),
+                };
+                values.push(value);
+            }
+        }
+
+        // Pad with zeros to reach power of 2
+        while values.len() < total_size {
+            values.push(M31(0));
+        }
```
ClementWalter, line 344: I don't like the fact that this is somehow dup code from the `write_trace`, can you propose something else?

crates/prover/src/preprocessed/mod.rs
```diff
@@ -55,6 +56,16 @@ impl PreProcessedTraceBuilder {
         self
     }
 
+    pub fn with_bitwise(mut self, operand_bits: u32) -> Self {
+        // Create the bitwise preprocessed columns
+        // This adds 4 columns: operation_id, input1, input2, result
+        let bitwise = bitwise::Bitwise::new(operand_bits);
```
ClementWalter, line 62: add `use crate::preprocessed::bitwise::Bitwise;` like for `RangeCheck`

PR #300, opened 2025-09-04
fix(compiler): resolve variable shadowing issues in semantic analysis

enitrat:
## Summary
- Fixed variable shadowing bugs where later variable declarations incorrectly affected earlier usages
- Refactored semantic index to properly track definition order and scope resolution
- Re-enabled previously disabled shadowing test cases

## Problem
The compiler had a bug where variable shadowing wasn't handled correctly. When a variable was redefined with a different type (e.g., from `felt` to `u32`), the type checker would incorrectly apply the later type to earlier usages of the variable, causing spurious type errors.

Example that was failing:
```cairo-m
let my_arr = [1, 2, 3];
let one = 1;           // felt type
let my_arr_1 = my_arr[one];  // Should use felt 'one'
let one = 1u32;        // Shadows previous 'one'
```

## Solution
Refactored the semantic index to:
- Use per-scope name indices that track definitions in order (oldest to newest)
- Remove PlaceTable in favor of direct name-to-definition mapping
- Track used definitions and identifier expression mappings
- Fix scope resolution to respect declaration order when handling shadowing

## Changes
- Updated semantic index structure and resolution logic
- Fixed MIR lowering to correctly handle shadowed variables
- Added comprehensive test coverage for shadowing scenarios
- Re-enabled previously disabled test in mdtest

## Test Plan
- [x] Existing tests pass
- [x] New shadowing test cases added and passing
- [x] Previously disabled mdtest now works correctly
- [x] Manual testing with various shadowing scenarios

Close CORE-1202


enitrat:
@codex review this PR. Make sure the code is clean, no dead code and dead comments.

PR #302, opened 2025-09-04
feat(prover): add u32_store_bitwise_fp_fp

AntoineFONDEUR:
closes CORE-1208


crates/prover/src/components/opcodes/u32_store_bitwise_fp_fp.rs
```diff
@@ -0,0 +1,831 @@
...
+//! - dst_prev_clock_lo
+//! - dst_prev_clock_hi
+//!
+//! # Constraints
+//!
+//! * enabler is a bool
+//!   * `enabler * (1 - enabler)`
+//! * opcode_constant is 36, 37 or 38
+//!   * `(36 - opcode_constant) * (37 - opcode_constant) * (38 - opcode_constant)`
+//! * define `bitwise_op = opcode_constant - 36`
```
ClementWalter, line 42: it is really necessary to add the constraint on the opcode_constant - 36/37/38 ? i mean that you can just extract the bitwise_op = opcode_constant - 36 and if it's not a valid value the lookup will fail

crates/prover/src/components/opcodes/u32_store_bitwise_fp_fp.rs
```diff
@@ -0,0 +1,831 @@
...
+
+        // Constraint: enabler is boolean
+        eval.add_constraint(enabler.clone() * (enabler.clone() - one.clone()));
+
+        // Constraint: opcode_constant is 36, 37, or 38
+        eval.add_constraint(
+            (opcode_constant.clone() - E::F::from(BaseField::from(U32_STORE_AND_FP_FP)))
+                * (opcode_constant.clone() - E::F::from(BaseField::from(U32_STORE_OR_FP_FP)))
+                * (opcode_constant.clone() - E::F::from(BaseField::from(U32_STORE_XOR_FP_FP))),
+        );
```
ClementWalter, line 620: I don't think that this is required, see above

ClementWalter (dismissed):

PR #305, opened 2025-09-05
feat(wasm): add basic memory operations

gael-bigot:
This PR adds support for:
- `i32.load`
- `i32.store`
- `global.get`
- `global.set`

### Method
- load and store operators added to MIR crate
- a "heap" is created at the end of VM memory which contains globals and allows for load and store operations

### Missing
- globals can't be initialized with a default value. This would require either adding a 'globals' field to Cairo-M programs or a main function
- several memories not supported

### Secondary features
- new instructions module to handle WASM -> MIR logic, split into `i32` and `memory`
- add support for void functions


enitrat:
@codex review this in depth

enitrat (commented):
Here's a small summary of what changed on main regarding interactions with direct memory:

  - New MIR Load/Store instructions that are _place_-based:
      - `InstructionKind::Load { dest, place, ty }`, `InstructionKind::Store { place, value, ty }`
      - A Place is MIR’s abstraction for an addressable memory location (an l-value). It describes “where” to load from or store to without embedding absolute addresses or raw
  offsets in MIR. A projection represents a small path of steps to reach a sub-location from the base, for example when a memory location represents the field of an aggregate
  - New MIR instruction for heap allocation `InstructionKind::HeapAllocCells { dest, cells } `
      - Pointers are represented as `MirType::Pointer { element: Box<MirType> }`
  - The underlying codegen lowers those instructions to CASM via double-derefs and a use of a global heap cursor that stores the top of the heap in a fixed memory cell.

How to go forward with this PR:
1. Keep changes in this PR restricted to memory operations so that the diff is clean, changes are scoped to only the new feature, and easier to review
2. There should not be a need to modify MIR/CASM crates beyond using the existing APIs.
3. To integrate with the MIR crate:
  - Do not push absolute addresses into MIR. Lower WASM memory to MIR pointers and Places. Only manipulate valid `ValueIds` or `Value` obtained from instructions
  - Allocate memory via HeapAllocCells, and use Load/Store on Places (pointer base + index).
  - Ensure MIR types are set correctly: pointer bases must be typed as MirType::pointer(elem_ty), loads/stores must pass the element type (e.g., MirType::U32).

4. For globals and i32Store / i32Load,
  - When encountered from the first time, you should allocate enough size on the heap and keep track of the pointer to that address, for example with a map`addresses_ptr: HashMap<usize, ValueId>` in the lowering context, so that you can map the WASM address to the corresponding MIR pointer
  - Replace the current “absolute address” with pointer-based globals.

PR #309, opened 2025-09-07
feat(semantic): forbid const arrays escaping by pointer (calls, aggregates, returns)

enitrat:
This PR adds semantic validation to prevent const arrays from escaping by pointer. It surfaces clear diagnostics early and avoids accidental mutation of read‑only const arrays via aliasing.

Summary
- Rejects passing const arrays to function parameters of type `[T; N]`.
- Rejects embedding const arrays into aggregates (struct fields and tuple elements) of type `[T; N]`.
- Rejects returning const arrays where the function return type is `[T; N]`.
- Adds a dedicated diagnostic code and thorough tests with snapshots.

Diagnostics
- Add `ConstArrayByPointer` (2017) to error when a const array is passed/embedded/returned by pointer.
  - File: `crates/compiler/diagnostics/src/diagnostics.rs`

Semantic checks (TypeValidator)
- Helper: detect when an expression resolves to a const array identifier (local or imported).
  - `expr_resolves_to_const_array(...)`
  - File: `crates/compiler/semantic/src/validation/type_validator.rs`

- Function calls: disallow passing const arrays to array parameters.
  - `check_function_call_types(...)` adds an error when `param_type` is `FixedArray` and arg resolves to a const array.
  - File: `crates/compiler/semantic/src/validation/type_validator.rs`

- Struct literals: disallow embedding const arrays in fields with array type.
  - `check_struct_literal_types(...)` adds an error for array field types whose value resolves to a const array.
  - File: `crates/compiler/semantic/src/validation/type_validator.rs`

- Tuple literals: disallow embedding const arrays in tuple elements.
  - `check_expression_types(...)` match arm for `Expression::Tuple` adds an error if any element resolves to a const array.
  - File: `crates/compiler/semantic/src/validation/type_validator.rs`

- Returns: disallow returning const arrays for functions with array return type.
  - `check_return_types(...)` adds an error when expected return type is `FixedArray` and the return expression resolves to a const array.
  - File: `crates/compiler/semantic/src/validation/type_validator.rs`

Tests
- New tests to validate behavior and diagnostics:
  - `crates/compiler/semantic/tests/arrays/const_arrays.rs`
    - `test_const_arrays_cant_be_written_to` (safety baseline)
    - `test_const_arrays_blocked_in_calls_and_aggregates` (calls, struct, tuple, return)
  - Snapshots:
    - `crates/compiler/semantic/tests/common/snapshots/parameterized__arrays::const_arrays::test_const_arrays_cant_be_written_to.snap`
    - `crates/compiler/semantic/tests/common/snapshots/parameterized__arrays::const_arrays::test_const_arrays_blocked_in_calls_and_aggregates.snap`

Rationale
- Arrays are passed by pointer; const arrays live in read‑only memory. Allowing these pointers to escape enables mutations through aliases. Enforcing this rule at the semantic layer gives users clear, actionable errors and prevents subtle bugs.

Follow‑ups (optional)
- Provide a standard `array_copy([T; N]) -> [T; N]` helper/builtin and reference it in diagnostics to guide remediation.
- If desired later: relax rule for functions proven read‑only on the parameter via simple analysis.

### DIFF
--- a/crates/compiler/diagnostics/src/diagnostics.rs
+++ b/crates/compiler/diagnostics/src/diagnostics.rs
@@ -120,6 +120,8 @@ pub enum DiagnosticCode {
     AssignmentToConst,
     IndexOutOfBounds,
     TypeInferenceError,
+    /// Passing or embedding a const array by pointer (disallowed); user must copy first
+    ConstArrayByPointer,
     // TODO: Add more type-related diagnostic codes:
     // - InvalidTypeAnnotation
     // - TypeArgumentMismatch
@@ -184,6 +186,7 @@ impl From<DiagnosticCode> for u32 {
             DiagnosticCode::AssignmentToConst => 2014,
             DiagnosticCode::IndexOutOfBounds => 2015,
             DiagnosticCode::TypeInferenceError => 2016,
+            DiagnosticCode::ConstArrayByPointer => 2017,
             DiagnosticCode::InternalError => 9001,
         }
     }
--- a/crates/compiler/semantic/src/validation/type_validator.rs
+++ b/crates/compiler/semantic/src/validation/type_validator.rs
@@ -102,6 +102,55 @@ impl Validator for TypeValidator {
 }
 
 impl TypeValidator {
+    /// Returns Some((name, name_span)) if the expression resolves to a const array
+    fn expr_resolves_to_const_array(
+        &self,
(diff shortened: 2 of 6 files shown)


enitrat:
@codex review

enitrat:
@codex review

PR #334, opened 2025-09-25
feat(wasm): add load and store ops

gael-bigot:
This PR adds support for the `i32.load` and `i32.store` opcodes in the wasm crate.
We assume a single infinite memory.
Addresses are converted as follows : `cm_addr = MAX_ADDRESS - wasm_base/2 - wasm_off/2 - 1`
`MAX_ADDRESS - wasm_off/2 - 1` is statically computed at transpile time, while `wasm_base/2` is computed dynamically using 2 MIR instructions (cast + div).
With the remaining sub and load/store, each i32.load/i32.store operation maps to 4 MIR instructions.

Test case `load_store.wat` checks both base addresses and offsets work correctly, and uses values above 2**16 to fill both limbs of u32 and detect any overlapping.

Snapshots show addresses 2^28-2, 2^28-4, 2^28-6... are being written to which is the expected behavior.

Closes CORE-1281


crates/wasm/src/lowering/ops.rs
```diff
@@ -186,7 +227,35 @@ impl DagToMirContext {
...
+                let result_id = self.mir_function.new_typed_value_id(MirType::U32);
+                let place = Place::new(cm_address);
+                let instruction = Instruction::load(result_id, place, MirType::U32);
+                self.get_current_block()?.push_instruction(instruction);
+                Ok(Some(result_id))
+            }
+
+            // Store I32 in memory
+            // See above for address computation
+            Op::I32Store { memarg, .. } => {
```
enitrat, line 251: https://rsms.me/wasm-intro#addressing-memory

@codex explain how the `memarg.align` flag behaves when our underlying memory is 16-bytes addressable

enitrat (commented):
looks good!

crates/wasm/src/lowering/ops.rs
enitrat: no, review this

enitrat (approved):

